PR #6, opened 2015-12-29
Automate adding CoreOS signing key to gpg public key ring

SpencerBrown:
and trusting it

this makes running get-coreos script more automated, no need for a manual step to import the CoreOS signing key


dghubble:
Merged as 3a915af9cdd85016f92f678b10331b6f99bdb0e9

PR #7, opened 2015-12-29
docs: updated links to virtual-hardware.md

sciphre:
Old filename was virtual-networking.md, which does not exist.


joshix:
LGTM; fixes a bug as yet unfiled.

PR #8, opened 2015-12-30
doc: Rename directory docs to Documentation

joshix:
Match the loose convention from other coreos projects, viz:
`https://github.com/coreos/^(etcd flannel rkt fleet)^/tree/master/Documentation/`

My personal preference is for ye olde unixism like @crawford's `ignition/doc/`.
However, if there is any current convention, it is for `Documentation`. This
changeset switches this repo to that convention. Either this, or let's discuss
switching all to `$project/doc/`. One way or another, let all repos do the same
thing -- except the coreos/docs/ repo itself, of course. :innocent:


joshix:
This PR may itself be a bug against the [style guide](https://github.com/coreos/docs/blob/master/STYLE.md). Perhaps add a note about standard doc location, once established.

dghubble:
Yeah, there does seem to be inconsistency here. It seems like most use `Documentation`, crawford uses `doc`, and https://github.com/coreos/clair uses `docs` like `coreos-baremetal`.

I'd be happy to change to `Documentation`. I'm not sure if coreos-pages setups expect all files under `Documentation` to be suitable for display on https://coreos.com. For example, does switching to `Documentation` mean https://github.com/coreos/coreos-baremetal/blob/master/docs/config.md needs to move since its not useful for display on coreos.com. None of the docs in this repo are on coreos.com we can move now and worry about that detail later perhaps.

joshix:
@dghubble The selection of pages for coreos.com publication is not recursive of the Doc^[...][s] directories in projects nor otherwise really automated. So, no, you don't need to divide docs intended for publication from those that are not.

dghubble:
Rebased and merged in 09a61b154f148f4bdaca8c51bbb48c42fbccbf54

PR #13, opened 2016-01-02
Parameter for build script to allow building for other than Linux

SpencerBrown:
builds binary as `server.<architecture>` e.g. `server.linux`, `server.darwin`
creates symlink from `server` to the last built binary

### DIFF
--- a/build
+++ b/build
@@ -1,5 +1,13 @@
 #!/bin/bash -e
 
+# check whether build architecture specified, otherwise default to linux
+GOOS="linux"
+if [ -n "$1" ]; then
+  GOOS="$1"
+fi
+
+echo "Building bin/server for $GOOS"
+
 # allow builds outside the standard GOPATH via symlink
 export GOPATH=${PWD}/Godeps/_workspace
 export GOBIN=${PWD}/bin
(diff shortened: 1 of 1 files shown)


dghubble:
The `build` script automates building the linux binary copied into the container image used locally or built for publishing to quay. If you'd like to build a Windows or OSX executable, the `run` script can be used directly. Is there use/value in writing out Windows/OSX binaries?

SpencerBrown:
I had looked at the `run` script but it doesn't save the binary. My use case is running `coreos-baremetal` and `pixiecore` running directly on OS X, which works for me in testing. In our company we have an OS X server that's on all the time, so it would be convenient to run PXE boots from that server.

[See here](https://github.com/SpencerBrown/coreos-pxe-install/blob/master/docs/coreos-baremetal.markdown) for my work-in-progress documentation. It's all manual steps now, and incomplete as to actually setting up a real working CoreOS cluster. But you can see where I'm going for the PXE boot piece.

dghubble:
Thanks for explaining your intention. I think, at least for now, `build` should just produce the binary suitable for use in the Docker image we publish. Folks wanting Linux, OSX, or Windows binaries can run `go build` themselves with whatever tags are appropriate. I'd just like to avoid `build` becoming a tool for general builds (with $CGO_ENABLED, $GOOS, $GOARCH, $TAGS etc.) when the go tools handle this well.

Also, symlinking bin/server to the latest binary means a Docker build may include the wrong one, depending on when it is done.

Something like the following should be fine for OSX unless you do care about cgo, debug, rebuild, and netgo flavor.

```
GOOS=darwin GOARCH=amd64 go build -o bin/server github.com/coreos/coreos-baremetal/cmd/bootcfg
```

SpencerBrown:
I'm fine with your conclusion, thanks for your time, attention, and clear explanation.

PR #14, opened 2016-01-04
Support a machines/default directory, for first time boot of new server

SpencerBrown:
The API server returns the Spec for machine "default" if it can't find a specific Spec for that machine.

```
data/
  machines/
    52:54:00:20:17:f9/
      machine.json
    default/
      machine.json
```

This allows me to boot new servers for the first time using a default configuration, and discover their MAC address so I can create a specific configuration.

### DIFF
--- a/api/spec.go
+++ b/api/spec.go
@@ -39,11 +39,15 @@ func (r *specResource) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 }
 
 // getMatchingSpec returns the Spec matching the given attributes.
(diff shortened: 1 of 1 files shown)


dghubble:
Booting new machines with a default definition to discover their attributes feels a bit dirty, but I was using that approach too (or just running `bootcfg` with `-log-level-=debug`). I agree this is of practical convenience for small clusters or random spare hardware whose spec sheets aren't readily available.

dghubble:
Feature added as #16

PR #22, opened 2016-01-12
cmd/bootcfg: Add version flag, set compile time version

dghubble:


cmd/bootcfg/main.go
```diff
@@ -13,38 +14,60 @@ import (
 	"github.com/coreos/pkg/flagutil"
 )
 
-var log = capnslog.NewPackageLogger("github.com/coreos/coreos-baremetal/cmd/bootcfg", "main")
+var (
+	// version provided by compile time flag: -ldflags "-X main.version $GIT_SHA"
+	version string
```
crawford: You might want something like https://github.com/coreos/ignition/blob/03fe11e0362873a7871f339c64b7d1d64365b08e/src/main.go#L38.

dghubble: Added the message, thanks. Also, the anonymous struct flags is a nice touch I've adopted it :+1:

PR #24, opened 2016-01-12
Add Kubernetes cluster example

dghubble:
* Add TLS asset generation based on coreos-kubernetes
* Add kubernetes example cluster (1 master, 1 worker, 1 etcd)

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -84,7 +84,7 @@ Reboot the nodes to PXE boot them into your new cluster!
 
 All examples use autologin so you can check whether your nodes were setup correctly, depending on the cluster example you chose.
 
-If something goes wrong, check troubleshooting.
+If something goes wrong, see [troubleshooting](../Documentation/troubleshooting.md).
 
 If everything works, congratulations! Stay tuned for developments.
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,33 +1,94 @@
 
 # Examples
 
-Examples contains `bootcfg` data directories showcasing different network-bootable baremetal clusters. These examples work with libvirt VMs setup with known hardware attributes by `scripts/libvirt`.
+Examples contains Config Service data directories showcasing different network-bootable baremetal clusters. These examples work with the libvirt VMs created by `scripts/libvirt` and the [Libvirt Guide](../Documentation/virtual-hardware.md).
 
-## Clusters
+| Name       | Description |  Docs          |
+|------------|-------------|----------------|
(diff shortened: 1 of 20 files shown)


examples/README.md
```diff
@@ -1,33 +1,62 @@
 
 # Examples
 
-Examples contains `bootcfg` data directories showcasing different network-bootable baremetal clusters. These examples work with libvirt VMs setup with known hardware attributes by `scripts/libvirt`.
+Examples contains Config Service data directories showcasing different network-bootable baremetal clusters. These examples work with the libvirt VMs created by `scripts/libvirt` and the [Libvirt Guide](Documentation/virtual-hardware.md).
```
aaronlevy: link to libvirt guide is broken

examples/README.md
```diff
@@ -1,33 +1,62 @@
...
+
+## Network Environment
+
+Run an iPXE setup with DHCP and TFTP on the virtual network on your machine similar to what would be present on a real network. This allocates IP addresses to VM hosts, points PXE booting clients to the config service, and chainloads iPXE.
+
+    docker run --rm --cap-add=NET_ADMIN coreos/dnsmasq -d -q --dhcp-range=172.17.0.43,172.17.0.99 --enable-tftp --tftp-root=/var/lib/tftpboot --dhcp-userclass=set:ipxe,iPXE --dhcp-boot=tag:#ipxe,undionly.kpxe --dhcp-boot=tag:ipxe,http://172.17.0.2:8080/boot.ipxe
+
+## Further Reading
+
+See the [libvirt guide](virtual-hardware.md) or [baremetal guide](physical-hardware.md) for more information.
```
aaronlevy: Both links broken

examples/kubernetes/README.md
```diff
@@ -0,0 +1,49 @@
+
+# Kubernetes
+
+This example provisions a Kubernetes cluster with 1 master node, 1 worker node, and a dedicated etcd node. Each node uses a static IP address on the local network.
+
+## Requirements
+
+First, follow the [examples](../README.md) documentation to setup client machines, the Config service, and the network environment.
```
aaronlevy: The ordering of these instructions are somewhat unclear:

The README says to look at the example after the config service, but before the network environment.
Then this says that it should be done after both the config service and the network environment. But per IRL discussion, the config service mount depends on the example being used.

May be easier to just have really linear instructions, even if it means duplicating some documentation per-example

aaronlevy:
After fixing links this LGTM. The instruction ordering is more of a nit / future thoughts if you just want to run with this.

PR #29, opened 2016-01-16
build: don't force GOOS=linux

philips:
There is no reason to force this. Instead have the docker-build command
do it before it is built.

Also, use the actual package name: bootcfg. It confused me that bootcfg
was called server.


philips:
I think we should remove the run script as it duplicates most of the build script logic. For example on etcd when iterating we just `./build && ./bin/etcd`.

PR #35, opened 2016-01-21
Add OpenPGP Signature Endpoints

dghubble:
Each config endpoint should have a sibling OpenPGP signature endpoint which returns the detached ascii armored signature of the HTTP response. (Dynamic) config responses are signed using an online private key (i.e. subkey) provided to the config server.


dghubble:
@gtank OpenPGP signature endpoints mentioned, fyi
@mjg59 no grub yet, but this can be made to handle the special case `.sig` for that endpoint

PR #66, opened 2016-02-24
examples: Show CoreOS install with Ignition support

### LINKED ISSUE #46: Remove grub.cfg write on install once coreos-install supports Ignition config
Currently, the install.service writes to grub.cfg to append kernel arguments to reference an Ignition config to use on the "first boot".

coreos-install support for Ignition will remove the need for this. Expected next OS release. https://github.com/coreos/init/pull/191

dghubble:
* Use the coreos-install -i option added in CoreOS alpha 949.0.0
* Download second stage Ignition config while in PXE first stage
* Resolves #46


crawford:
LGTM

Issue #71, opened 2016-02-24
Kubernetes v1.2.* Conformance tests


dghubble:
k8s conformance tests fail locally because they pull very large Docker images while `bootcfg` examples PXE boot CoreOS instances to run in RAM (edit: not longer true). We could allocate disks from a local pool, format them, boot from a root fs, and (optionally) install CoreOS to disk in an early Ignition state, before provisioning Kubernetes.

* Disks were added to the libvirt examples in fc4d669a386
* Use these disks for the root filesystem. Makes local nodes more practical for experimentation too
* https://coreos.com/kubernetes/docs/latest/conformance-tests.html

dghubble:
Against v1.2.2 on physical machines via coreos-baremetal.

```
[91m�[1mSummarizing 4 Failures:�[0m

�[91m�[1m[Fail] �[0m�[90mConfigMap �[0m�[91m�[1m[It] should be consumable from pods in volume as non-root [Conformance] �[0m
�[37m/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/util.go:1685�[0m

�[91m�[1m[Fail] �[0m�[90mConfigMap �[0m�[91m�[1m[It] should be consumable from pods in volume with mappings as non-root [Conformance] �[0m
�[37m/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/util.go:1685�[0m

�[91m�[1m[Fail] �[0m�[90mKubectl client �[0m�[0mKubectl logs �[0m�[91m�[1m[It] should be able to retrieve and filter logs [Conformance] �[0m
�[37m/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/kubectl.go:812�[0m

�[91m�[1m[Fail] �[0m�[90mSchedulerPredicates [Serial] �[0m�[91m�[1m[It] validates resource limits of pods that are allowed to run [Conformance] �[0m
�[37m/go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/test/e2e/util.go:2159�[0m

�[1m�[91mRan 95 of 270 Specs in 3145.215 seconds�[0m
[...]

PR #73, opened 2016-02-25
examples: Partition disk, root fs, and PXE with root fs

### LINKED ISSUE #72: examples: Partition disk, root fs, and PXE with the root fs
* By default, CoreOS PXE images run from RAM.

dghubble:
* Resolves #72, needed for #71


dghubble:
@crawford

Issue #77, opened 2016-03-02
Use /etc/bootcfg for configs by default


dghubble:
Default to /etc/bootcfg as the location for the top-level YAML config, ignition configs, cloud configs, and profiles. This behavior is more conventional and friendlier for running directly as a host application (systemd).

PR #102, opened 2016-03-15
cmd, docs: Change default config, data, and assets locations

dghubble:
* Change default -config path to /etc/bootcfg.conf
* Change default -data-path to /etc/bootcfg
* Change default -assets-path to /var/bootcfg


dghubble:
Resolves #77

PR #109, opened 2016-03-17
examples: Fix 2-stage install CoreOS and etcd_proxy setup

dghubble:
* 4GB disk is too small and routinely fails in libvirt
* Depending on build env, dnsmasq ACI can have an old /etc/resolv.conf
DNS entry embedded. Can rebuild to fix, but use --dns=8.8.8.8 too
* Fix issue where etcd_proxy profile failed to pull metadata
so etcd_proxy wasn't configured after CoreOS install

### DIFF
--- a/Documentation/getting-started-rkt.md
+++ b/Documentation/getting-started-rkt.md
@@ -92,7 +92,7 @@ Build the `dnsmasq.aci` ACI.
 
 Run `dnsmasq.aci`.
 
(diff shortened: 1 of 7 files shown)


dghubble:
@gyuho @xiang90 FYI, https://github.com/coreos/coreos-baremetal/blob/master/Documentation/getting-started-rkt.md with the `coreos-install.yaml` config validates a 3-node etcd and proxy setup on bare metal.

Issue #103, opened 2016-03-15
Merge bootcfg HTTP server and gRPC server


dghubble:
Fold the cmd/bootcfg-rpc/main.go gRPC test server into the main bootcfg binary to start providing gRPC support. Decide whether to run the services on different ports or to multiplex.

dghubble:
For now, these will run on independent ports configured by `-address=127.0.0.0:8080` and `-rpc-address=127.0.0.1:8081`. Leaving off the RPC address flag means the experimental RPC server won't be started.

PR #110, opened 2016-03-17
cmd/bootcfg: Start HTTP and gRPC(optional) via bootcfg binary

dghubble:


dghubble:
cc @sym3tri

cmd/bootcfg/main.go
```diff
@@ -108,16 +113,35 @@ func main() {
...
+	})
+
+	// gRPC Server (feature hidden)
+	if flags.rpcAddress != "" {
+		grpcServer, err := rpc.NewServer(bootcfgServer)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Infof("starting bootcfg gRPC server on %s", "127.0.0.1:8081")
+		lis, err := net.Listen("tcp", "127.0.0.1:8081")
```
sym3tri: Why not listen on `flags.rpcAddress`?

sym3tri:
Otherwise changes LGTM.

I'd start looking into this... might save you some time.
https://coreos.com/blog/gRPC-protobufs-swagger.html

cmd/bootcfg/main.go
dghubble: oh thanks, that was the entire point of adding rpcAddress :)

dghubble:
With Go 1.6 we can multiplex HTTP and gRPC off the same port like in Brandon's example and I'll remove the optional `rpcAddress`. #103

Issue #100, opened 2016-03-11
Upgrade Kubernetes to v1.1.8 with CoreOS hyperkube, rkt kubelet


dghubble:
* https://github.com/coreos/coreos-kubernetes/pull/276
* https://github.com/coreos/coreos-kubernetes/pull/302

PR #112, opened 2016-03-19
examples: Upgrade k8s to v1.1.8

dghubble:
* Upgrade CoreOS Version to Alpha 962.0.0
* rkt run kubelet with the kubelet-wrapper
* Switch to quay.io/coreos/hyperkube
* Provision an ext4 root partition


dghubble:
Solves #100.
@aaronlevy

aaronlevy:
lgtm

Issue #60, opened 2016-02-16
Migrate Kubernetes examples from cloud-config to Ignition


dghubble:
* k8s-rkt.yaml
* k8s-docker.yaml

crawford:
![LGTM](http://media.giphy.com/media/CcngBIJfWunn2/giphy.gif)

PR #115, opened 2016-03-21
Rewrite k8s-master and k8s-worker Profiles from cloud-config to Ignition

dghubble:


dghubble:
Verified Kubernetes PXE and Kubernetes Install with rkt and docker on libvirt machines. Verified Kubernetes PXE and Kubernetes Install on real hardware as well (via rkt).

dghubble:
fyi @crawford @aaronlevy

dghubble:
After a k8s-install to disk, the cluster no longer requires metadata from bootcfg on reboots. However, the k8stls.service unit runs on each reboot (noticed after an auto-update). If bootcfg isn't around any more (ephemeral provisioner - i.e. my laptop) the kubelet unit must be manually modified and started (not ok).

The credentials written by k8stls.service are fetched since they cannot be part of the machine metadata (typically version controlled). This k8stls.service really only needs to run once ever, so one approach might be to disable itself at the end or make itself a NoOp so the kubelet can continue. Or another way to achieve the same effect.

examples/ignition/k8s-master.yaml
```diff
@@ -0,0 +1,551 @@
...
+        WantedBy=multi-user.target
+
+storage:
+  {{ if .pxe }}
+  disks:
+    - device: /dev/sda
+      wipe_table: true
+      partitions:
+        - label: ROOT
+          number: 0
```
crawford: This can be omitted. If you want to be explicit, use `number: 1`.

examples/ignition/k8s-master.yaml
```diff
@@ -0,0 +1,551 @@
...
+          contents: |
+            {
+              "apiVersion": "v1",
+              "kind": "Namespace",
+              "metadata": {
+                "name": "kube-system"
+              }
+            }
+        - path: /opt/init-flannel
+          mode: 320
```
crawford: This is an odd mode. Since this is YAML, you can specify this in octal to make it easier to read.

examples/ignition/k8s-worker.yaml
```diff
@@ -0,0 +1,182 @@
...
+        WantedBy=multi-user.target
+
+storage:
+  {{ if .pxe }}
+  disks:
+    - device: /dev/sda
+      wipe_table: true
+      partitions:
+        - label: ROOT
+          number: 0
```
crawford: Same as above.

examples/ignition/k8s-master.yaml
dghubble: "mode must be properly specified as a decimal value" Are the docs incorrect? I converted to decimal. https://coreos.com/ignition/docs/latest/configuration.html

dghubble:
Resolves #60

examples/ignition/k8s-worker.yaml
dghubble: Removed.

examples/ignition/k8s-master.yaml
dghubble: Uses 0500, but maybe 0644 would be more conventional / less paranoid.

examples/ignition/k8s-master.yaml
dghubble: Removed

examples/ignition/k8s-master.yaml
dghubble: Using 0644

examples/ignition/k8s-master.yaml
```diff
@@ -0,0 +1,559 @@
...
+        [Path]
+        PathExists=/etc/kubernetes/ssl/ready
+    - name: kubelet.service
+      enable: true
+      contents: |
+        [Unit]
+        Description=Kubelet via Hyperkube ACI
+        Requires=flanneld.service
+        After=flanneld.service
+        Requires=kubelet.path
```
crawford: These aren't needed. The path unit will trigger the service for you.

examples/ignition/k8s-master.yaml
```diff
@@ -0,0 +1,559 @@
...
+        WantedBy=multi-user.target
+    - name: kubelet.path
+      enable: true
+      contents: |
+        [Unit]
+        Description=Watch for Kubelet TLS Assets
+        [Path]
+        PathExists=/etc/kubernetes/ssl/ready
+    - name: kubelet.service
+      enable: true
```
crawford: Probably don't want to enable this.

examples/ignition/k8s-master.yaml
```diff
@@ -0,0 +1,559 @@
...
+        RestartSec=10
+        [Install]
+        WantedBy=multi-user.target
+    - name: k8s-addons.service
+      enable: true
+      contents: |
+        [Unit]
+        Description=Start Kubernetes DNS Controller and Service
+        Requires=kubelet.service
+        After=kubelet.service
```
crawford: This is going to end up triggering kubelet.service before it's ready to go.

examples/ignition/k8s-master.yaml
dghubble: Er, 0544

examples/ignition/k8s-master.yaml
dghubble: kubelet.path removed

examples/ignition/k8s-master.yaml
dghubble: Replaced by a `Requires=` and `After=` on `k8s-assets.target`

examples/ignition/k8s-master.yaml
dghubble: I've left this enabled now that the path unit isn't used

crawford:
LGTM

PR #117, opened 2016-03-24
bootcfg/storage: Switch fileStore from http.Dir to custom Dir

dghubble:
* http.Dir securely restricts accesses to a filesystem directory to
support http FileServe, which is why this type was used originally
* Dir copies the sanatization behavior of http.Dir verbatim, but
provides ioutil-like write and directory access within the directory
* Why: Safe FileStore writes need to be possible and also, storage
should not have a dependency on net/http just for its http.Dir

### DIFF
--- a/bootcfg/storage/file_store.go
+++ b/bootcfg/storage/file_store.go
@@ -1,0 +1,92 @@
+package storage
+
+import (
+	"encoding/json"
+	"path/filepath"
+
+	"github.com/coreos/coreos-baremetal/bootcfg/storage/storagepb"
+)
+
+// Config initializes a fileStore.
+type Config struct {
+	Dir    string
+	Groups []*storagepb.Group
+}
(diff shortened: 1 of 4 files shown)


bootcfg/storage/file_store.go
```diff
@@ -0,0 +1,92 @@
...
+// Config initializes a fileStore.
+type Config struct {
+	Dir    string
+	Groups []*storagepb.Group
+}
+
+// fileStore implements ths Store interface. Queries to the file system
+// are restricted to the specified directory tree.
+type fileStore struct {
+	root   string
```
dghubble: This file is just a move except for lines concerning the root tree node.

dghubble:
@gtank

dghubble:
Fixed "sanitize" spelling. Thx

PR #120, opened 2016-03-28
Refactor Group and Profile selection into core server

dghubble:
* Add gRPC SelectGroup and SelectProfile service endpoints
* HTTP and gRPC servers should both allow labels to be resolved to the matching Group or Profile. Move logic to bootcfg core server
* Add gRPC wrapper selectServer to avoid cluttering core server with response types which are currently only useful for gRPC


dghubble:
@sym3tri

bootcfg/api/handlers.go
```diff
@@ -25,3 +30,37 @@ func logRequests(next http.Handler) http.Handler {
 	}
 	return http.HandlerFunc(fn)
 }
+
+// selectGroup selects the Group whose selectors match the query parameters,
+// adds the Group to the ctx, and calls the next handler. The next handler
+// should handle a missing Group.
```
sym3tri: nit: having comments here is great for GoDoc. It's also nice to have them in the proto file so the entire API is human parsable with 1 file. Maintaining 2 sets of identical comments is not good, so I propose putting them in just the proto file b/c the actual server implementation methods are unlikely to be dependencies of anything.

sym3tri: disregard. i thought this was the rpc handler.

dghubble: hm, selectGroup is a HTTP handler that deals with query parameters, contexts, and http.Requests. The proto service SelectGroup is the underlying logic which is shared.

Maybe within the server package I'll leave comments in the proto, but in `api` (probably should be called `www` or `web`) they'd make sense?

bootcfg/server/serverpb/rpc.proto
```diff
@@ -19,6 +19,27 @@ service Profiles {
   rpc ProfileList(ProfileListRequest) returns (ProfileListResponse) {};
 }
 
+service Select {
+  rpc SelectGroup(SelectGroupRequest) returns (SelectGroupResponse) {};
```
sym3tri: What does "Selection" actually do?

dghubble: There should be several methods like these. SelectGroup, SelectProfile, SelectIgnition, SelectCloudConfig. Basically machine labels (usually from query parameters) are used to select the first matching Group, Profile, Ignition Config, etc.

bootcfg/rpc/select.go
```diff
@@ -0,0 +1,30 @@
...
+
+func newSelectServer(s server.Server) pb.SelectServer {
+	return &selectServer{
+		srv: s,
+	}
+}
+
+func (s *selectServer) SelectGroup(ctx context.Context, req *pb.SelectGroupRequest) (*pb.SelectGroupResponse, error) {
+	group, err := s.srv.SelectGroup(ctx, req)
+	return &pb.SelectGroupResponse{Group: group}, err
```
sym3tri: Maybe in a separate PR...
it's best to use standard gRPC error codes with your returned errors.

https://godoc.org/google.golang.org/grpc#Errorf

bootcfg/api/handlers.go
sym3tri: i think it's fine as-is. sorry my first comment was a mistake.

bootcfg/server/serverpb/rpc.proto
sym3tri: WFM. Maybe brief comments would be good here.

sym3tri:
otherwise LGTM

bootcfg/rpc/select.go
dghubble: Ok, I'll do that in a followup soon

dghubble:
Thanks

PR #121, opened 2016-03-28
Add storage/testutils to share fake Storage implementations

dghubble:


dghubble:
@ericchiang Since we discussed test packages yesterday

PR #123, opened 2016-03-28
examples: Remove metadata unit from etcd examples

dghubble:
The etcd clusters use static address assignment so dynamic metadata isn't needed for this case. Custom metadata is useful for fetching custom data on each boot, to configure units.


dghubble:
Rebased onto master directly, intermediate commits aren't needed.

PR #126, opened 2016-03-29
scripts: Simplify TLS scripts into k8s-certgen and generate kubeconfig

dghubble:
* Remove kubecfg-rkt and kubecfg-docker which used relative paths
* Generate a kubeconfig with embedded certs at the same time TLS certs are created. This kubeconfig is "portable" since it does not rely on relative paths. Also eliminates the confusion of having one for docker and rkt networks.


dghubble:
@aaronlevy

scripts/tls/kube-conf
```diff
@@ -0,0 +1,44 @@
...
+}
+
+if [ -z "$1" ] || [ -z "$2" ]; then
+    usage
+    exit 1
+fi
+
+DEST="$1"
+MASTER_IP="$2"
+ADMIN_CERT_BASE64=$(cat $DEST/admin.pem | base64 -w 0)
```
aaronlevy: Dunno if it matters but osx base64 has no `-w` flag so this will fail.

scripts/tls/kube-conf
```diff
@@ -0,0 +1,44 @@
...
+- name: coreos-baremetal
+  cluster:
+    certificate-authority-data: ${CA_CERT_BASE64}
+    server: https://${MASTER_IP}:443
+contexts:
+- context:
+    cluster: coreos-baremetal
+    user: coreos-baremetal
+  name: coreos-baremetal
+current-context: coreos-baremetal
```
aaronlevy: nit: having all items named the same could cause confusion. The general [upstream approach](http://kubernetes.io/docs/user-guide/kubeconfig-file/) is to suffix with `-context` `-user` `-cluster`

aaronlevy:
Couple minor comments, but otherwise lgtm

scripts/tls/kube-conf
dghubble: done

scripts/tls/kube-conf
dghubble: removed on darwin

PR #128, opened 2016-03-31
examples: Add a POC of how AWS Ignition serving would work

dghubble:
* Ignition v2.0.0 could fetch the Ignition config from bootcfg
* Sets up the etcd nodes using the discovery endpoint

### DIFF
--- a/examples/groups/etcd-aws/default.json
+++ b/examples/groups/etcd-aws/default.json
@@ -1,0 +1,15 @@
+{
+	"id": "etcd-aws",
+	"name": "etcd Node",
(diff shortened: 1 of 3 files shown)


dghubble:
For now, validate with something like. Later instance can query bootcfg.
    
    ignition=$(curl "http://bootcfg.foo/ignition?name=etcd-aws&platform=aws")
    aws ec2 run-instances --image-id=ami-e1720f81 --count=3 --instance-type=t2.nano --subnet-id=mysubnet --associate-public-ip-address --user-data="${ignition}"

PR #138, opened 2016-04-07
scripts: get-coreos pulls a clean, current alpha by default

joeatwork:
The proposed change automatically pulls image assets from /current rather than a pinned version. That means that maintainers won't have to chase versions to keep the script up to date.

The change requires that the get-coreos script clean up before pulling new assets, since current doesn't have a unique name and a mix of old and new assets could end up in the pulled directory if the script or internet connection failed in the right way.

### DIFF
--- a/scripts/get-coreos
+++ b/scripts/get-coreos
@@ -3,7 +3,7 @@
 # USAGE: ./scripts/get-coreos channel version
 
 CHANNEL=${1:-"alpha"}
-VERSION=${2:-"983.0.0"}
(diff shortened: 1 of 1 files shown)


crawford:
There is a coupling between the version of the OS and the config that is provided. There is going to need to be a bit more version-skew detection logic to make this robust.

joeatwork:
Closing this, it turns out we'll need to be a bit more deliberate about how this works.

PR #143, opened 2016-04-12
examples: Upgrade k8s from v1.1.8 to v1.2.0

dghubble:
Validated both iPXE booted clusters and Installed to Disk clusters using `bootcfg` running via rkt (i.e. `k8s` and `k8s-install` examples). Tested with Tectonic Console and some `Deployments`. Resolves #122 

@pbx0 @aaronlevy

Issue #142, opened 2016-04-11
Add Heapster to the Kubernetes example profile


dghubble:
Convert this patch set to Ignition and pull in any recent changes https://github.com/coreos/coreos-kubernetes/commit/dbb5bf4480ce24f241a50cd668fdaad47b6b7410

PR #144, opened 2016-04-12
Start Heapster by default in k8s examples

dghubble:
@pbx0 @joeatwork


dghubble:
Checked Heapster is working on clusters while doing the k8s v1.2.0 upgrade in #143. Solves issue #142

PR #171, opened 2016-04-26
docs: Add rkt, Docker, and Kubernetes bootcfg deployment notes

### LINKED ISSUE #93: docs: Deployment via Kubernetes

dghubble:
* Resolves #93

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -19,6 +19,24 @@ Run the latest or the most recently tagged `bootcfg` [release](https://github.co
 
 Create machine profiles, groups, or Ignition configs at runtime with `bootcmd` or by using your own `/var/lib/bootcfg` volume mounts.
 
+## Kubernetes
+
+*Note: Enhancements to the gRPC API, CLI, and `EtcdStore` backend will improve this deployment strategy in the future.*
+
+Create a `bootcfg` Kubernetes `Deployment` and `Service` based on the example manifests provided in [contrib/k8s](../contrib/k8s).
+
+    kubectl apply -f contrib/k8s/bootcfg-deployment.yaml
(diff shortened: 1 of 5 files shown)


dghubble:
@colhom Does infra have a NodePort selection policy? I've used  32401 in the `bootcfg` example, a random-ish choice. Asking because you chose a port in https://github.com/coreos/docker-nginx-https-redirect/blob/master/contrib/k8s/https-redirect-service.yaml#L11

colhom:
@dghubble no real rule for assigning node ports. I tend to use 31XXX for production services, and 32XXX for their staging analogues. Just a "pattern", not really a rule.

I expect that we'll be transitioning away from nodeports as we can make better use of load-balancer services anyways. Maintaining this mapping out-of-band is hideous.

dghubble:
I'll just use the randomly selected 31488 NodePort for the generic example then. Folks who have k8s compatible load balancers can edit how they like. Thanks Colin!

chancez:
BTW, you should be able to not specify the node port, and k8s will pick for you, alternatively, I'd actually say you should just leave it at `type: ClusterIP` as the default.

dghubble:
Hm, maybe I've optimized for my typical use cases. I choose a fixed NodePort per service globally (rather than letting k8s decide) to support load balancing across nodes in multiple Kubernetes clusters (baby ubernetes). The Service IP is determined at the load balancer level.

What's your aim for using ClusterIP in the example?

chancez:
Well, I guess using a clusterIP doesnt make much sense, since nodes wont be on the k8s network, thus they wouldn't be able to use the clusterIP to access it. Never mind.

joeatwork:
+1 for fixed ports in examples

PR #175, opened 2016-04-26
docs: Remove cloud-config from machine lifecycle diagram

dghubble:
* coreos-baremetal no longer uses examples with cloud-config
so the diagram can be simplified to just show Ignition


dghubble:
@crawford

PR #177, opened 2016-04-28
contrib/systemd: Run bootcfg with bootcfg user and group

dghubble:
* Setup /var/lib/bootcfg with the bootcfg group

Deployment opinions from @colhom or @cgag?


dghubble:
rkt has a script for setting up groups and the data dir. Maybe we want something similar. https://github.com/coreos/rkt/blob/master/dist/scripts/setup-data-dir.sh

Issue #176, opened 2016-04-26
Stop requiring Ignition templates to specify their rendered format


joeatwork:
JSON templates are really hard to write and maintain, so maybe we can remove support for them? Please close this if it sounds goofy, or comment and I'll remove JSON support.

dghubble:
I left support for JSON Ignition templates for users who wish to use their own conversion tools to produce Ignition JSON. For example, someone might convert YAML files to JSON manually with a tool like [fuze](https://github.com/coreos/fuze) or maybe they prefer toml or some other format `bootcfg` doesn't support. On the other hand, I'm not aware of anyone relying on this.

I don't know that I'd call this a feature request. More of a removal.

joeatwork:
Here's the feature associated with the request - accept YAML over the API without specifying the type, or using an id named `$SOMETHING.yaml` - it's definitely not a show stopper, there are other solutions (for example, we could try to parse an expanded template as both flavors, we could allow API clients to tell us explicitly what type their payload was, or we could just treat ids of the form `$SOMETHING.yaml` specially) but it seems like whenever you can accomodate a use case by cutting code/making an API simpler, it's a nice way to go.

On the other side of things, I'm not sure that fuze/toml/what-have-you is useful in our general case since the templates aren't legal YAML or legal JSON. In my case, I can't read and parse YAML and produce JSON, or generate JSON from programming language literals, because of constructs like the following in my template:

```
{{ if .ssh_authorized_keys }}
passwd:
  users:
    - name: core
      ssh_authorized_keys:
        {{ range $element := .ssh_authorized_keys }}
        - {{$element}}
        {{end}}
{{end}}
```

Let's talk about this when we can :)

dghubble:
Yeah, presently the file type is determined by the file extension. Perhaps the proposed `IgnitionTemplate` in #166 takes the file type or maybe it can be reliably inferred from the content.

The use case for fuze/toml/etc. is simply that the conversion to JSON is done out-of-band so a user can provide a raw legal JSON they've prepared however. Running it through the template is a NoOp so then it is just validated to be valid Ignition JSON. How valuable that flexibility is to people, I don't know.

You're right, in the standard case users are providing template contents (perhaps `.tmpl`) which render to YAML or JSON. They're not YAML or JSON upfront, so I agree it is weird those file extensions must be used.

joeatwork:
Discussed - going to go with (something like)
```
		// Unmarshal YAML or JSON Ignition config
[...]

Issue #141, opened 2016-04-11
Support for Ignition v0.4.0 and config version "2.0.0"


dghubble:
- [x] Allow V2 Ignition YAML
- [x] Allow V2 Ignition JSON
- [x] Migrate Example Profiles ([migration guide](https://github.com/coreos/fuze/blob/master/doc/configuration.md))

Bumped out:
* Respect Ignition Headers (bump to #248, serving machines Ignition 2.0.0 works for higher versions)
* Translate from V1 to V2 if requested (no longer relevant, Fuze YAML does not specify a version)

dghubble:
Blocked by https://github.com/coreos/bugs/issues/1259

dghubble:
The plan is for YAML components of Ignition to be split out into https://github.com/coreos/fuze. Humans will write (or template with bootcfg) fuze configs (a YAML document). Fuze will formalize the transform between a fuze config and different Ignition JSON documents for machines to read.

The changes to `bootcfg` "ignition" templates will be minimal (but breaking) as `bootcfg` has done a similar transform on its own for quite a while. A WIP of these changes can be found in the [fuze-configs](https://github.com/coreos/coreos-baremetal/commits/fuze-configs) branch. This unblocks the migration to the Ignition 2.0.0 format for all example clusters.

If this change is ready, it would be great to include in the coreos-baremetal v0.4.0 release.

cc @crawford

PR #178, opened 2016-04-28
ignition: Allow Ignition V1/V2, ignore template file extension

dghubble:
* Allow V1 and V2 Ignition JSON
* Allow V1 and V2 Ignition YAML
* Stop requiring Ignition templates to use a file extension to specify the rendered format (#176)
* Relates to #141


bootcfg/http/ignition.go
```diff
@@ -59,28 +62,70 @@ func ignitionHandler(srv server.Server) ContextHandler {
...
-	return strings.HasSuffix(filename, ".yaml") || strings.HasSuffix(filename, ".yml")
+// parseToV2 parses raw JSON in Ignition v2 format and returns the
+// Ignition v2 Config struct.
+func parseToV2(data []byte) (cfg ignitionTypes.Config, err error) {
+	// parse JSON v2 to Ignition
+	cfg, err = ignition.ParseFromLatest(data)
+	if err == nil {
+		return cfg, nil
+	}
+	switch majorVersion(data) {
```
joeatwork: why isn't this
```
if 2 != majorVersion(data) {
   err = yaml.Unmarshal(data, &cfg)
}
```

dghubble:
This change merely allows Ignition V1/V2 YAML/JSON to be used to begin migrating. Other Ignition 2.0.0 upgrade issues are tracked in #141.
cc @crawford

bootcfg/http/ignition.go
dghubble: Thanks, just a holdover from the auto-upgrade V1 -> V2 flow I was toying with, I'll change it to that.

PR #184, opened 2016-05-04
Remove 8.8.8.8 from networkd example configs

dghubble:
* Profiles should use only the user-provided DNS endpoint.
Adding 8.8.8.8 can cause inconsistent resolution and in some
setups, 8.8.8.8 is not routable.


crawford:
LGTM

Issue #192, opened 2016-05-10
Example PXE doesn't work now


gosharplite:
I have a working PXE example with CoreOS 899.17.0. The latest PXE example with CoreOS 1032.0.0 is not working. Using the same network environment, the differences are shown below.
```
$ diff pxe.json.OLD pxe.json.NEW
6,7c6,7
< 		"kernel": "/assets/coreos/899.17.0/coreos_production_pxe.vmlinuz",
< 		"initrd": ["/assets/coreos/899.17.0/coreos_production_pxe_image.cpio.gz"],
---
> 		"kernel": "/assets/coreos/1032.0.0/coreos_production_pxe.vmlinuz",
> 		"initrd": ["/assets/coreos/1032.0.0/coreos_production_pxe_image.cpio.gz"],
```
```
$ diff ssh.yaml.OLD ssh.yaml.NEW
2c2,3
< ignition_version: 1
---
> ignition:
>   version: 2.0.0
```
[Screenshot of CoreOS 899.17.0.](https://raw.githubusercontent.com/gosharplite/scrap/master/OLD.png) (working)
[Screenshot of CoreOS 1032.0.0.](https://raw.githubusercontent.com/gosharplite/scrap/master/NEW.png) (not working)

Network environment commands.
```
sudo docker run -p 8080:8080 --rm -v $PWD:/var/lib/bootcfg:Z -v $PWD/groups/pxe:/var/lib/bootcfg/groups:Z quay.io/coreos/bootcfg:v0.3.0 -address=0.0.0.0:8080 -log-level=debug
```
```
sudo docker run --net=host --rm --cap-add=NET_ADMIN quay.io/coreos/dnsmasq -d -q -i eth1 --dhcp-range=10.128.112.18,10.128.112.24 --enable-tftp --tftp-root=/var/lib/tftpboot --dhcp-userclass=set:ipxe,iPXE --dhcp-boot=tag:#ipxe,undionly.kpxe --dhcp-boot=tag:ipxe,http://bootcfg.foo:8080/boot.ipxe --log-queries --log-dhcp --dhcp-option=3,10.128.83.254 --address=/bootcfg.foo/10.128.112.12
```
VM creation command.
```
virt-install --name tony-node1 -u 16e7d8a7-bfa9-428b-9117-363341bb330b --pxe --disk pool=default,size=6 --boot=hd,network --network=bri
dge:vif.112 --ram=1024 --vcpus=1 --os-type=linux --noautoconsole
```

[...]

PR #195, opened 2016-05-10
example: Revert ssh Ignition config change

dghubble:
* Users try to use the latest reference example clusters with
a bootcfg v0.3.0 deployment, which did not yet support Ignition
2.0.0. Wait until closer to v0.4.0 to bump Ignition configs.


dghubble:
Closes #192

PR #198, opened 2016-05-10
docs/getting-started-rkt: the --insecure-options flag was unnecessary

dgonyeo:
The docs had the user skip image verification for an image pulled from
Quay, claiming that since Quay hosts docker images the image is
unsigned. This is true when fetching the docker file with "docker://",
but when an ACI is pulled from Quay (which it converts on the fly), the
ACI will be properly signed.


crawford:
LGTM

dghubble:
As long as folks understand that a Quay signature of the latest build is not the same as the `coreos.com/bootcfg` ACI signed by the CoreOS App key.

PR #204, opened 2016-05-16
examples/k8s: Update k8s to v1.2.4 and openssl conf

dghubble:
* Add missing kubernetes.default.svc alt name to openssl config

cc @aaronlevy @chancez

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,7 +10,7 @@
 
 #### Examples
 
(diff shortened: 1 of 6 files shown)


chancez:
LGTM.

aaronlevy:
If you bump it to `v1.2.4_coreos.1` it has git and nfs-common as part of the image ....

dghubble:
That works too. Bumped.

PR #217, opened 2016-05-26
support Darwin in Makefile

SpencerBrown:
@dghubble how does this look to you? I'm heading towards running bootcfg etc. directly on OS X, and testing with VirtualBox.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,18 @@
+LOCAL_OS := $(shell uname)
 
-BIN_DIR=/usr/local/bin
-DATA_DIR=/var/lib/bootcfg
-ENV_FILE=/etc/bootcfg.env
+ifeq ($(LOCAL_OS),Linux)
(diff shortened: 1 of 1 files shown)


dghubble:
You said you're testing locally and `~/.bootcfg` seems like a fine choice if it works for you. Since the Makefile is used for building and installing the source release for the deployment [docs](https://github.com/coreos/coreos-baremetal/blob/master/Documentation/deployment.md) we'd want that to choose config and data directories for users to run a cross compiled bootcfg in production.

On Linux, the `/var/lib/boocfg` default and `/etc/bootcfg` default locations were chosen since they're the conventional service data and env locations according to the Linux FHS. For OSX, I'm not positive what the conentional/good-citizen system directories would be.

We were planning to include a pre-compiled binary for darwin next release so OSX is something we will want to think about.

PR #218, opened 2016-05-27
contrib/dnsmasq: CAP_NET_BIND_SERVICE is needed

artem-sidorenko:
to allow socket binding

https://github.com/coreos/rkt/issues/2706


dghubble:
Merged within #220

PR #219, opened 2016-05-27
Script to install on local OSX

SpencerBrown:
I'm looking to support testing bare metal CoreOS locally on OS X using coreos-baremetal, dnsmasq, and VirtualBox. 

#217 modified Makefile to install `bootcfg` on OS X. By @dghubble comments, I understand Makefile is for building production environments and so he does not want to merge that PR.

Instead I offer this script, to configure/install `bootcfg` locally on OS X. This is to be run after the `build` script is run to build the Darwin compatible binary.

### DIFF
--- a/build-osx
+++ b/build-osx
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -e
+
+BIN_DIR=/usr/local/bin
+DATA_DIR=~/.bootcfg
+ENV_FILE=~/.bootcfg/env
+
+if [[ -e $DATA_DIR ]]
+then echo "Data directory ${DATA_DIR} already exists"
+else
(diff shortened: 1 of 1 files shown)


chancez:
You could just `export GOOS=darwin` and run everything as normal.

dghubble:
I'm not sure I see the need. The data directory setup was recently moved out of the Makefile and into the [setup-data-dir](https://github.com/coreos/coreos-baremetal/blob/master/scripts/setup-data-dir) script so a script makes more sense. Wouldn't this script be specific to your setup though with `~/.bootcfg`? Someone else on OSX might want `~/.config/bootcfg` or something.

If you run `make` and `make install` that should build and install `bootcfg` on OSX since `GOOS=darwin` should be set already. The examples use `examples` and `examples/assets`, but OSX users can choose custom locations for developing locally according to their preferences.

PR #221, opened 2016-05-27
*: Use agnostic Makefile, stop creating empty env file

dghubble:
* systemd users can use /etc/bootcfg.env if they choose to use environment files


SpencerBrown:
perfect. obviates both my PRs, I will close them and rework.

dghubble:
Great, thanks for bringing it up!

PR #225, opened 2016-06-01
*: an (old) procfile for running libvirt/rkt examples and tests

joeatwork:
This is for @dghubble's reference, but probably shouldn't be merged in it's originally proposed form.


dghubble:
Thanks for sharing this Joe. The Procfiles are nice, but I'm just not sure how best to integrate them into examples while still showing users how everything works (their main goal). I thought about putting them in contrib too, but I'm worried about maintaining them. https://github.com/coreos/coreos-baremetal/blob/master/Documentation/config.md has several commands already.

For now I'll close this, but feel free to open if there are other ideas.

PR #232, opened 2016-06-08
Log error when profile not found

SpencerBrown:
I found this useful when debugging iPXE booting.

@dghubble if you like this direction I can add more logging to the other handlers. If you want this to be debug level logging vs. error I can do that too.

### DIFF
--- a/bootcfg/http/ipxe.go
+++ b/bootcfg/http/ipxe.go
@@ -34,6 +34,7 @@ func ipxeHandler() ContextHandler {
 	fn := func(ctx context.Context, w http.ResponseWriter, req *http.Request) {
 		profile, err := profileFromContext(ctx)
 		if err != nil {
+			log.Errorf("Error finding profile: %v", err)
 			http.NotFound(w, req)
 			return
 		}


dghubble:
Sure, how about logging these in the corresponding spots in grub, ipxe (done), pixiecore, cloud, and ignition too? This is well timed, we'll be switching the logging library soon so good to get this in place.

rothgar:
Is this a good place to ask for more logging information? I have been wanting more than just the HTTP GET request url when debug logging is enabled.
It be great for debug to show the URL but then also show what group, profile, and ignition/cloud config was read. I often have to do this manually and sometimes still run into issues tracking down why something is loading.
I can open a separate issue for this functionality unless @dghubble you will be adding this with the logging library switch.

dghubble:
@rothgar can you put your feature requests in #233?

dghubble:
Also, debug seems more suitable here than error.

dghubble:
Closing in favor of switching to Logrus #254 and adding structured logging #268 (examples shown). Feel free to re-open if I've missed a concern.

PR #238, opened 2016-06-14
scripts: update rkt-gc-force for rkt 1.8.0

sym3tri:
the previous script was not working for me with rkt 1.8.0

### DIFF
--- a/scripts/rkt-gc-force
+++ b/scripts/rkt-gc-force
@@ -7,6 +7,7 @@ for p in `mount | grep -oP "/var/lib/rkt/pods/[^\s,\)]*"`; do umount -l $p; done
 rkt gc --grace-period=0s
 for i in `rkt image list | grep -oP '^sha512[^ \t]*'`; do rkt image rm $i; done
 for ns in `ip netns l | grep -o -E '^[[:alnum:]]+'`; do ip netns del $ns; done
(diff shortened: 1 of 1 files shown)


dghubble:
Did you see the update to https://github.com/coreos/coreos-baremetal/issues/208? In theory, the rkt-gc-force script would only be needed for rkt 1.6 and below, before the netns changes were made. I guess it depends if this still happened if rkt v.1.8 was the first version you've installed and it still happens?

I'm not sure what the netns naming was before, we'd probably want to keep the line that removed the pre-v1.6.0 netns's?

sym3tri:
Need to consult with the rkt team for ideas.

FWIW 1.8.0 was the 1st and only version I ever installed on this machine.

dghubble:
Hmm, that's not a promising sign that this is actually fixed.

sym3tri:
No it's not.

Since each of those lines fail silently we could modify the script to do BOTH the old command and the new one. Admittedly a heavy-handed unattractive approach.

dghubble:
Yeah, let's do that. If its a nuclear clean script, might as well clean for both versions.

dghubble:
This doesn't seem to be enough

dghubble:
Wrong button

sym3tri:
@dghubble did that

Issue #224, opened 2016-05-31
Add /kickstart endpoint for rendering ks files


adamschaub:
To make coreos-baremetal more relevant in heterogeneous environments. Adding a /kickstart endpoint will allow non-CoreOS machines to make the same use of group metadata that ignition/cloud endpoints do.

brianredbeard:
This sounds completely reasonable.  Due to some of the intricacies of how Anaconda parses Kickstart files and the expectations around automatically handling LVM configuration, packages, etc I would say we should identify someone with this type of Anaconda expertise and have them collaborate.

adamschaub:
@brianredbeard Point well taken. Any recommendations? I hear you're a Red Hat survivor.

brianredbeard:
@adamschaub While we've tried to not paint anything into a corner on this, we're also making sure that it allows for a first class deployment experience for CoreOS hosts.  Before we can spend any time worrying about the experience for rpm based distros we need to get it right for ourselves.  This project was borne out of a number of deficiencies (for the deployment of CoreOS at least) in Cobbler, Spacewalk, Foreman, Razor, Juju, etc.  When folks from those communities belly up to the bar to collaborate we'll be happy to have them along for the ride, but we can't spare the cycles at the moment to work on non-core features at the expense of direct needs that I've had for a number of years.

adamschaub:
@brianredbeard Couldn't agree more, I don't want to push to derail current priorities. We can provide the development cycles. To reaffirm, if we bring in external effort to contribute this feature, it could find a home in the project?

dghubble:
The project structure should allow booting and provisioning other distributions, it just has CoreOS config file support at the moment. I'd like to see kickstart (or other) configs supported, but don't have the time to spend on it right now so contributions would be welcome.

One option might be to support a `/generic` config endpoint which just hooks into the group matching and profiles and templates any files provided without further logic. This means folks could try out serving kickstart files (or other configs), but without type or validation support until that's ready and can move to a dedicated, named endpoint. Its enough to try it out and see how it feels.

adamschaub:
Sorry for the issue ref clutter. I realized late into the process that I needed to wrestle a few files around to make cleaner commits, and bad decisions ensued O.o

I've issued a PR implementing a `/generic` endpoint for the bootcfg HTTP server, as @dghubble had suggested. Let me know how you feel about the changes. Some feature like this could really help us fit coreos-baremetal into our architecture.

dghubble:
[...]

PR #239, opened 2016-06-14
bootcfg/{http,server,storage}: Adding a generic endpoint

adamschaub:
Adding a `/generic` endpoint to allow rendering arbitrary go templates from bootcfg. This PR is related to the discussion in #224.


bootcfg/http/generic.go
```diff
@@ -0,0 +1,62 @@
...
+		// collect data for rendering
+		data := make(map[string]interface{})
+		if group.Metadata != nil {
+			err = json.Unmarshal(group.Metadata, &data)
+			if err != nil {
+				log.Errorf("error unmarshalling metadata: %v", err)
+				http.NotFound(w, req)
+				return
+			}
+		}
```
dghubble: You can add the same special value `data["query"] = req.URL.RawQuery` that is provided to Ignition templates. Its acceptable here because people will use generic for experimentation and edge cases.

bootcfg/storage/storagepb/storage.proto
```diff
@@ -25,8 +25,10 @@ message Profile {
   string ignition_id = 3;
   // cloud config id
   string cloud_id = 4;
+  // generic config id
+  string generic_id = 5;
   // support network boot / PXE
-  NetBoot boot = 5;
+  NetBoot boot = 6;
```
dghubble: Changing protocol buffer numeric tags breaks gRPC API backwards compatability. Although I don't know if external users depend on the gRPC API (which will be documented and TLS auth'd with v0.4.0), we use it internally and its best practice to use the next number for new values. See https://developers.google.com/protocol-buffers/docs/proto3 "Updating a Message Type".

adamschaub: Ah, thanks for the reference. Makes sense in retrospect :) My base instinct was to group the id's.

bootcfg/http/generic.go
adamschaub: Great stuff. I'll rope rope it in.

dghubble:
Overall this looks pretty good. Can you add a 3rd commit

* Add a line to `scripts/setup-data-dir`
* Include generic in https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootcfg.md#data
* Add a genericID to https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootcfg.md#profiles to hint at that this is possible. I'd hold off on having a dedicated page for generic configs until there is an example.

Those are the places in the docs off the top of my head, there may be other spots. I'll check this out and test it on something a bit later.

Possibly in a separate PR:

* Add an `examples/generic` directory.
* Add a minimal example with a machine profile that references a generic template.

adamschaub:
@dghubble Thanks for the quick turn around! I've addressed your inline comments, plus added the documentation where you've indicated. 

I'll definitely set up an example. I'll plan to submit this in a separate PR.

dghubble:
Great, thanks! Rebased and merged.

After trying it out, I can maybe see changing "generic" to "other" before the next release. But let's let it simmer for now. Only thing, in future can you wrap commit lines before 80?

adamschaub:
> Only thing, in future can you wrap commit lines before 80?

Sure thing. Thanks for reviewing!

PR #240, opened 2016-06-15
bootcfg: pass client hostname/ip from ignition http request header to ignition template

chancez:


dghubble:
We could maybe add it to the `.request` template variable namespace later, though we both talked about it being odd to rely upon that IP value. For `v0.4.0` let's leave it out, this PR doesn't compile / pass and its a bit out of date.

PR #241, opened 2016-06-15
bootcfg/http: pass query params to templates in ignition

chancez:


chancez:
I'd like to namespace these variables under a subkey like `http`, but not all variables exposed today offer this, which makes it a bit odd. If I had my way: selectors would be accessed under `{{ .group.selector.mySelector }}` and these variables would be under `{{ .http.query_params.myQueryParam }}` but I'm not sure we want to break existing templates to get to better namespacing.

dghubble:
Query params are made available in https://github.com/coreos/coreos-baremetal/pull/278 as `{{.request.query.foo}}`.

PR #243, opened 2016-06-16
examples/bootkube: Use bootkube image and disable auto-updates

dghubble:
* Users no longer need to scp a bootkube binary
* Disable auto-updates until kube-apiserver checkpointing is done

@aaronlevy


aaronlevy:
lgtm

PR #252, opened 2016-06-28
Makefile: Add release tarball target

dghubble:
@sym3tri

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,18 @@ uninstall:
 	rm $(BIN_DIR)/bootcfg
 	rm $(BIN_DIR)/bootcmd
 
-.PHONY: build test install
+release: clean _output/coreos-baremetal-linux-amd64.tar.gz
+
+_output/coreos-baremetal-%-amd64/bootcfg: build
+	@mkdir -p $(dir $@)
+	./scripts/release-files $(dir $@)
+
+_output/coreos-baremetal-%-amd64.tar.gz: _output/coreos-baremetal-%-amd64/bootcfg
+	tar zcvf $@ -C _output coreos-baremetal-$*-amd64
+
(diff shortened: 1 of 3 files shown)


sym3tri:
lgtm

Makefile
```diff
@@ -22,4 +22,18 @@ uninstall:
...
+
+_output/coreos-baremetal-%-amd64.tar.gz: _output/coreos-baremetal-%-amd64/bootcfg
+	tar zcvf $@ -C _output coreos-baremetal-$*-amd64
+
+clean:
+	rm -rf _output
+
+.PHONY: build clean install test
+
+.SECONDARY: _output/coreos-baremetal-linux-amd64/bootcfg
```
crawford: I think you actually want `.INTERMEDIATE` here. That way this gets cleaned up.

Makefile
```diff
@@ -22,4 +22,18 @@ uninstall:
 	rm $(BIN_DIR)/bootcfg
 	rm $(BIN_DIR)/bootcmd
 
-.PHONY: build test install
+release: clean _output/coreos-baremetal-linux-amd64.tar.gz
+
+_output/coreos-baremetal-%-amd64/bootcfg: build
+	@mkdir -p $(dir $@)
```
crawford: I would use an order-only prerequisite to create the directory.

dghubble: Ok, added in https://github.com/coreos/coreos-baremetal/pull/258

Makefile
dghubble: Make tries to remove the target intermediate directory (and fails `unlink: _output/coreos-baremetal-linux-amd64: Is a directory`), so SECONDARY was just a way to avoid that and ignore it.

PR #264, opened 2016-07-11
Makefile,scripts: Make platform-specific binary tarball targets

dghubble:


Makefile
```diff
@@ -1,44 +1,45 @@
...
 
-_output/coreos-baremetal-%-amd64:
-	mkdir -p $@
+bin/%/bootcmd:
+	GOOS=$* go build -o bin/$*/bootcmd -ldflags $(LD_FLAGS) -a github.com/coreos/coreos-baremetal/cmd/bootcmd
 
-_output/coreos-baremetal-%-amd64.tar.gz: bin/bootcfg bin/bootcmd | _output/coreos-baremetal-%-amd64
-	./scripts/release-files $|
-	tar zcvf $@ -C _output coreos-baremetal-$*-amd64
+_output/coreos-baremetal-%-amd64.tar.gz: NAME=coreos-baremetal-$(VERSION)-$*-amd64
```
crawford: Why bother with pattern-specific variables? You can just set these in the recipe.

dghubble: We talked and decided it was better as is than to add within the recipe

    $(eval NAME=coreos-baremetal-$(VERSION)-$*-amd64)
    $(eval DEST=_output/$(NAME))

PR #281, opened 2016-07-19
scripts: Don't set GPG trust level of CoreOS public key

### LINKED ISSUE #280: scripts: get-coreos should not set ultimate trust
IMO, its a bad practice to add our public key to a user's GPG keyring with a trust level (ultimate). Users run this script in production to fetch new PXE image releases so we shouldn't muck around with trust levels.

Following this change, running `get-coreos` imports the CoreOS public key into the user's gpg keyring, but **does not** define a trust level.

## Mini Tutorial in GPG Trsut

Users should cross-reference public keys and determine their own level of trust going forward. Uninitiated GPG users who leave the trust level "unknown" should expect to see valid signatures, but ambiguous trust of the key.

```
gpg: Good signature from "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>"
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
```

Once you've consulted with all the trusted parties in all the lands and feel confident that the GPG key you've imported does indeed belong to CoreOS, you may set a trust level.

```
gpg --edit-key 04127D0BFABEC8871FFB2CCE50E0885593D2DCB4
trust
# input trust level
quit
gpg --list-keys
```

dghubble:
cc @gtank 

Closes #280

### DIFF
--- a/scripts/get-coreos
+++ b/scripts/get-coreos
@@ -1,6 +1,7 @@
-#!/bin/bash -e
+#!/bin/bash
 # USAGE: ./scripts/get-coreos
 # USAGE: ./scripts/get-coreos channel version dest
+set -eou pipefail
 
 CHANNEL=${1:-"alpha"}
 VERSION=${2:-"1053.2.0"}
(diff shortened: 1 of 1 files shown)


gtank:
LGTM

PR #282, opened 2016-07-20
Added CloudPut methods/tests for client/rpc/storage/server

mikeynap:
Working on a project with Cisco and we needed the ability to add CloudConfig files over gRPC. 

Is this something that you would be comfortable allowing?


mikeynap:
Which version of protoc is this using?

bootcfg/storage/testfakes/fixtures.go
```diff
@@ -39,12 +39,23 @@ var (
 
 	// IgnitionYAMLName is an Ignition template name for testing.
 	IgnitionYAMLName = "ignition.tmpl"
-
+
```
danehans: ^ looks like whitespace that needs to be removed.

bootcfg/storage/testfakes/fixtures.go
```diff
@@ -39,12 +39,23 @@ var (
...
+	
 	// IgnitionYAML is an Ignition template for testing.
 	IgnitionYAML = `ignition_version: 1
 systemd:
   units:
     - name: etcd2.service
       enable: true
 `
+	// CloudYAMLName is an CloudConfig template name for testing.
+	CloudYAMLName = "ignition.tmpl"
```
danehans: should this be cloud.tmpl to differentiate from the ignition template?

dghubble:
Although Ignition is recommended, the `/cloud` endpoint is still supported and gRPC API coverage is in-scope. I'll take a look through after `v0.4.0` is squared away this week.

dghubble:
I believe protoc-3.0.0-beta-2. This needs to be captured in the build process.

dghubble:
I've tested `make tools` and `make codegen` against your PR. Once https://github.com/coreos/coreos-baremetal/pull/305 merges you should be able to run those re-generate code and tests should pass.  Thanks for highlighting the lack of a pinned protocol buffer build target.

PR #290, opened 2016-07-25
examples/etcd: Use DNS names in etcd clusters, no IPs

dghubble:
* Use DNS names to refer to nodes in etcd examples to mirror production
* Add dnsmasq.conf files for metal0 (rkt) and docker0 examples which include static MAC->IP and Name->IP mappings
* Remove the etcd-docker example cluster, no longer needed

cc @gyuho - changes the bare metal reference cluster for etcd.
rel #291


gyuho:
lgtm.  Defer to @xiang90 

Thanks.

gyuho:
Oh never mind. It belong to baremetal. Thanks!

Issue #291, opened 2016-07-25
Migrate examples to use DNS names only, no IPs


dghubble:
Reference clusters should use DNS names for all nodes, not IP addresses.

* Mirrors production setups
* DNS is effectively a requirement for multi-master Kubernetes setups
* Differences between local VM clusters on metal0 (rkt) and docker0 can be abstracted
* Easier to understand. Doc readers understand that `node1.example.com` could be anything appropriate  for a particular environment, whereas, 172.15.0.21 feels special.

rel. #261

dghubble:
Question about flanneld which came up: https://github.com/coreos/flannel/issues/470. Its not a hard requirement to solve this issue.

dghubble:
Migration of self-hosted is blocked by https://github.com/coreos/bootkube/issues/89

dghubble:
Choices for using DNS names instead of IP names with self-hosted Kubernetes now seem to be forking bootkube, waiting for https://github.com/kubernetes/kubernetes/pull/27880 upstream, or adding that patch to the CoreOS hyperkube so bootkube can make use of the new node name downward api.

dghubble:
Example clusters use DNS names to refer to fellow members (peers, workers, etc.) rather than IPs.

PR #292, opened 2016-07-25
examples/k8s: Use DNS names in Kubernetes clusters

dghubble:
* Use DNS names to refer to nodes to mirror production

@aaronlevy 
rel #291, https://github.com/dghubble/metal/commit/3fcb8829ba9c4d1744ac4de5a3704ba9de36f848


examples/README.md
```diff
@@ -46,7 +46,3 @@ Most examples allow `ssh_authorized_keys` to be added for the `core` user as mac
 ### "pxe"
 
 Some examples check the `pxe` variable to determine whether to create a `/dev/sda1` filesystem and partition for PXEing with `root=/dev/sda1` ("pxe":"true") or to write files to the existing filesystem on `/dev/disk/by-label/ROOT` ("pxe":"false").
-
-### "skip_networkd"
```
dghubble: We no longer set a fixed networkd config, no need to explicitly skip.

Documentation/kubernetes.md
```diff
@@ -15,9 +15,8 @@ Ensure that you've gone through the [bootcfg with rkt](getting-started-rkt.md) o
 
 The [examples](../examples) statically assign IP addresses to libvirt client VMs created by `scripts/libvirt`. VMs are setup on the `metal0` CNI bridge for rkt or the `docker0` bridge for Docker. The examples can be used for physical machines if you update the MAC/IP addresses. See [network setup](network-setup.md) and [deployment](deployment.md).
 
-* [k8s](../examples/groups/k8s) - iPXE boot a Kubernetes cluster (use rkt)
```
dghubble: People who still use docker can just use `k8s` and `k8s-install` directly. They work for rkt and docker.

examples/ignition/k8s-master.yaml
```diff
@@ -196,7 +195,6 @@ storage:
       filesystem: rootfs
       contents:
         inline: |
-          FLANNELD_IFACE={{.ipv4_address}}
```
dghubble: Flannel uses the interface of the default route by default. https://github.com/coreos/flannel/issues/470

examples/ignition/k8s-master.yaml
```diff
@@ -157,7 +157,6 @@ storage:
               - --allow-privileged=true
               - --service-cluster-ip-range={{.k8s_service_ip_range}}
               - --secure-port=443
-              - --advertise-address={{.ipv4_address}}
```
dghubble: Removing this flag should mean the address bound to the default interface is advertised. http://kubernetes.io/docs/admin/kube-apiserver/

examples/ignition/k8s-worker.yaml
```diff
@@ -58,7 +58,7 @@ systemd:
           --register-node=true \
           --allow-privileged=true \
           --config=/etc/kubernetes/manifests \
-          --hostname-override={{.ipv4_address}} \
+          --hostname-override={{.domain_name}} \
```
dghubble: Any name which resolves to the host can be used here. We discussed some variable name choices (including `name`, `dns_name`), but decided it was better to go with `domain_name`. `name` is too broad and `dns_name` implied the metadata value was used for DNS configuration when it is not.

dghubble:
Merged as dc5b3e24e5e774f245caf4894807aa9dface73dc

PR #294, opened 2016-07-26
examples/torus: Use DNS names for Torus cluster

dghubble:
* Also change `torusblk volume create` to `torusctl block create`

cc @barakmich 
rel #291


dghubble:
Merged as 4d468484171c7ad8ff97900fb557d24eaa4b6fb6

PR #296, opened 2016-07-27
examples: Update kubelet.service to mount /var/log in rkt

### LINKED ISSUE #288: Default to mounting /var/log in kubelet-wrapper
https://coreos.com/kubernetes/docs/latest/kubelet-wrapper.html#use-the-cluster-logging-add-on

I've helped quite a few people who are using coreos-baremetal who've ran into issues with clustered logging. We already mount in the host's `resolv.conf`, maybe in addition we should default to mounting in `/var/log` so the logging addon works without any changes.

@dghubble what do you think?

chancez:
Fixes #288

### DIFF
--- a/examples/ignition/bootkube-master.yaml
+++ b/examples/ignition/bootkube-master.yaml
@@ -44,7 +44,7 @@ systemd:
         Description=Kubelet via Hyperkube ACI
         Wants=flanneld.service
         [Service]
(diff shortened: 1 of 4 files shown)


dghubble:
Rebased as 8705f78aee6d2a50a83df94c7b8d2cfcaf931253

chancez:
SGTM

PR #301, opened 2016-08-04
Updated DNS values in kubernetes-cert script

### LINKED ISSUE #299: Update k8s-certgen DNS.*=kubernetes* so they don't conflict with user flags
If a user tries to create certs and uses DNS.[1-4]= they will conflict and override the existing kubernetes* names used in the script. This can cause problems later depending on what name they use to discover the kubernetes server

I tried making the script smarter (aka, ugly `awk`) by reading the DNS number and incrementing it by 4 but realized  that wouldn't always work if someone has `DNS.1=foo,DNS.5=bar` or if someone provides values out of order `DNS.5=foo,DNS.2=bar,DNS.1=baz`

The best option is probably to use large DNS.# values like @dghubble suggested on slack or move to a smarter tool like [cfssl](https://github.com/cloudflare/cfssl).

Also related to #291

rothgar:
I just added 100 to each existing DNS value which worked in my test.

fixes #299

### DIFF
--- a/scripts/tls/kubernetes-cert
+++ b/scripts/tls/kubernetes-cert
@@ -43,10 +43,10 @@ keyUsage = nonRepudiation, digitalSignature, keyEncipherment
 subjectAltName = @alt_names
 
 [alt_names]
-DNS.1 = kubernetes
-DNS.2 = kubernetes.default
-DNS.3 = kubernetes.default.svc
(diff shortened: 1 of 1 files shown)


dghubble:
Rebased and merged as db9141ec1ebadce72543b100a51b42a4e5b4e7bf

Issue #269, opened 2016-07-12
Document installation on CoreOS


dghubble:
Overhaul the installation docs, with a dedicated section for those using CoreOS as their provisioning host. The recommended systemd unit file should use `rkt`, unlike the existing `contrib/systemd` example which makes use of the binary.

PR #306, opened 2016-08-05
Documentation: Add guide for installing on CoreOS

dghubble:
* Provide a systemd service unit which rkt runs bootcfg

rel. #269 
cc. @sym3tri

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -5,37 +5,106 @@ This guide walks through deploying the `bootcfg` service on a Linux host (via bi
 
 ## Provisoner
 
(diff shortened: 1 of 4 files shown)


Documentation/deployment.md
```diff
@@ -5,80 +5,125 @@ This guide walks through deploying the `bootcfg` service on a Linux host (via bi
...
+$ tar xzvf coreos-baremetal-v0.4.0-linux-amd64.tar.gz
+$ cd coreos-baremetal-v0.4.0-linux-amd64
+```
+
+## TLS Credentials
 
-### Copy Binary
+*Skip this unless you need to enable the gRPC API*
+
+The `bootcfg` gRPC API allows external client tools (`bootcmd` CLI, graphical apps) to manage how machines are provisioned so TLS credentials are needed for client authentication and to establish a secure communication channel. Client machines read from the HTTP endpoints and do not require this setup.
```
sym3tri: "Client machines" here is a bit vague. Maybe be more explicit as to what type of clients. 

i.e. "Client machines which are PXE booting"

PR #307, opened 2016-08-05
scripts: Move node info into common sourceable scripts

chancez:


scripts/common.sh
```diff
@@ -0,0 +1,15 @@
+NODE1_NAME=node1
+NODE1_UUID=16e7d8a7-bfa9-428b-9117-363341bb330b
```
dghubble: We don't require fixed UUID's for examples anymore. I'd remove that bit.

dghubble:
Just the one cleanup, then looks good.

scripts/common.sh
chancez: Do they cause any problems? I think it could still be useful to have consistent UUIDs, even if they're not used in examples.

dghubble: No, but if not required I don't want to pin them. Same reason we don't pin a serial id and other unused things.

chancez:
@dghubble Updated

PR #309, opened 2016-08-09
Add ARM to release architectures

dghubble:
I've been playing with `bootcfg` on some ARM boards and other folks might like to do the same. Include it in the next release.

cc @crawford Makefile opinions?

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -96,9 +96,11 @@ $ sudo systemctl start bootcfg.service
 
 ### General Linux
 
-Pre-build binaries are available for general Linux distributions. Copy the `bootcfg` static binary to an appropriate location on the host.
+Pre-built binaries are available for general Linux distributions.
 
 #### Binary
+
+Copy the `bootcfg` static binary to an appropriate location on the host.
 
 ```sh
 $ sudo cp bootcfg /usr/local/bin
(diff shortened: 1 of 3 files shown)


urzds:
Does this mean that `contrib/systemd/matchbox-for-tectonic.service` (and hence the [Tectonic tutorial](https://tectonic.com/enterprise/docs/latest/install/bare-metal/index.html)) works unmodified on ARM systems, like the RasPi? Which instruction set (`GOARM=...`) does this target? ARMv6 (RasPi 1) or ARMv7 (RasPi 2+)?

dghubble:
In the Makefile, `GOARM` is unset, which I believe defaults to 6. It would be better if we were explicit if you wanted to submit something.

These arm builds would let you run matchbox on an ARM box as your provisioner if you wanted. We validate these clusters and tools in a lab of typical typical beefy enterprise servers of various sorts (mostly x86_64). So fair warning that ARM on these pi's and things is for fun at the moment.

urzds:
The [Go documentation](https://github.com/golang/go/wiki/GoArm) says "Starting from Go 1.1, the appropriate GOARM value will be chosen if you compile the program from source on the target machine." I understand this as meaning it works like `-march=native` on GCC.

The major problem with running Matchbox on RasPi is that currently there is no distribution which works out of the box: Rasbian supports ARMv6 (RasPi1), but [does not package Rkt](https://bugs.launchpad.net/raspbian/+bug/1655118), while Debian packages Rkt, but supports only ARMv7 (RasPi2+) and [requires](https://wiki.debian.org/RaspberryPi) some [manual twiddling](http://sjoerd.luon.net/posts/2015/02/debian-jessie-on-rpi2/) to [boot](https://github.com/drtyhlpr/rpi23-gen-image) on any generation of Raspberry Pi.

Until that is resolved, I don't think investing much time into supporting Tectonic/Matchbox on this system is a worthwhile endeavor.

dghubble:
If we're just talking about running the `matchbox` binary to serve configs to client machines on a network, there is no requirement the provisioner has rkt. Configuring client machines like Raspberry Pi or Beaglebone or others to PXE boot is out of scope and difficult/tedious depending on the board. As an aside CoreOS Container Linux is not supported on Raspberry Pi devices either.

Thanks for your take. Yep, Tectonic isn't spending time on Raspberry Pi support. Enterprises and datacenters are the focus.

urzds:
My idea was to use a RasPi as bootcfg machine.

By the way: is Tectonic arch agnostic? I.e. can I use it in a mixed ARM/ARM64/AMD64 environment and it will select the right binaries (i.e. matchbox, kubelet, apiserver, ...) depending on the host that is being provisioned?

PR #317, opened 2016-08-17
examples: Update self-hosted Kubernetes to v1.3.4

dghubble:
* Use bootkube v0.1.4 for self-hosted bootstrapping


dghubble:
Tested both `bootkube` and `bootkube-install`

PR #324, opened 2016-08-23
examples: Use universal root filesystem

dghubble:
* Use the "root" filesystem from the Ignition universal base config (path /sysroot)
* No need for custom named filesystem anymore
* https://coreos.com/ignition/docs/latest/examples.html#create-files-on-the-root-filesystem

wip: Update other reference clusters to replace "rootfs" with "root".
cc : @crawford


crawford:
LGTM

dghubble:
Changed "rootfs" to "root" for other clusters as well.

PR #328, opened 2016-08-25
Update and add to etcd3 example clusters

dghubble:
* Use notify service type so dependent services don't start immediately
* Add etcd3-install example cluster
    * etcd3-install installs CoreOS to disk and sets up a 3-node etcd3 cluster
    * Additional machines are setup as etcd3 proxies

cc @gyuho


gyuho:
lgtm. thanks!

PR #330, opened 2016-08-26
examples: Update example clusters to CoreOS 1153.0.0

dghubble:
* CoreOS 1153.0.0 adds rkt 1.13.0 which should resolve a docker2aci bug in rktnetes https://github.com/coreos/rkt/pull/3026

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -32,8 +32,8 @@ Finds the profile for the machine and renders the network boot config (kernel, o
 **Response**
 
     #!ipxe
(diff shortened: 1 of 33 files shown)


dghubble:
Install to disk examples aren't working https://github.com/coreos/bugs/issues/1541. They work with CoreOS 1109.1.0 and 1122.0.0. But not 1151.0.0.

dghubble:
Bumping to CoreOS Alpha 1153.0.0

Issue #331, opened 2016-08-26
get-coreos fails on HTTP/2 web-server response


urzds:
CoreOS webservers reply with `HTTP/2 200`, but get-coreos fails on that response.

```
# curl -s -I https://stable.release.core-os.net/amd64-usr/1068.9.0/coreos_production_pxe.vmlinuz
HTTP/2 200 
...
```

urzds:
I suggest following regex instead: `/^HTTP\/[0-9.]+ [23][0-9][0-9]$/`

dghubble:
Are you using an old copy of the script? Switched to HTTP/2 in https://github.com/coreos/coreos-baremetal/commit/39f5c526eceed76c7a3ee2cdd9336e96a1d17936

```
./scripts/get-coreos stable 1068.9.0
Creating directory REDACTED
Downloading CoreOS stable 1068.9.0 images and sigs to REDACTED
CoreOS Image Signing Key
######################################################################## 100.0%
gpg: key 50E0885593D2DCB4: "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
coreos_production_pxe.vmlinuz...
######################################################################## 100.0%
coreos_production_pxe.vmlinuz.sig
######################################################################## 100.0%
coreos_production_pxe_image.cpio.gz
######################################################################## 100.0%
coreos_production_pxe_image.cpio.gz.sig
######################################################################## 100.0%
coreos_production_image.bin.bz2
######################################################################## 100.0%
coreos_production_image.bin.bz2.sig
######################################################################## 100.0%
gpg: Signature made Thu 04 Aug 2016 10:18:41 PM PDT
gpg:                using RSA key 48F9B96A2E16137F
gpg: Good signature from "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>"
gpg: Signature made Thu 04 Aug 2016 10:18:42 PM PDT
gpg:                using RSA key 48F9B96A2E16137F
gpg: Good signature from "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>"
gpg: Signature made Thu 04 Aug 2016 10:06:32 PM PDT
gpg:                using RSA key 48F9B96A2E16137F
gpg: Good signature from "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>"
```

urzds:
[...]

PR #334, opened 2016-08-27
get-coreos: Be more relaxed about the channel-or-version-not-found check

devurandom:
Previously this matched very specific HTTP status codes only, while now it matches any success or redirection status code. It also works for "HTTP/2" answers in addition to "HTTP/2.0".

Fixes: #331

### DIFF
--- a/scripts/get-coreos
+++ b/scripts/get-coreos
@@ -10,7 +10,10 @@ DEST=$DEST_DIR/coreos/$VERSION
 BASE_URL=https://$CHANNEL.release.core-os.net/amd64-usr/$VERSION
 
 # check channel/version exist based on the header response
-curl -s -I $BASE_URL/coreos_production_pxe.vmlinuz | awk '/2.0 200/ {found++} /1.1 200/ {found++} /1.1 301/ {found++} END { if (found<1) { print "Channel or Version not found"; exit 1 }}'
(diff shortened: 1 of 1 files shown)


scripts/get-coreos
```diff
@@ -10,7 +10,10 @@ DEST=$DEST_DIR/coreos/$VERSION
 BASE_URL=https://$CHANNEL.release.core-os.net/amd64-usr/$VERSION
 
 # check channel/version exist based on the header response
-curl -s -I $BASE_URL/coreos_production_pxe.vmlinuz | awk '/2.0 200/ {found++} /1.1 200/ {found++} /1.1 301/ {found++} END { if (found<1) { print "Channel or Version not found"; exit 1 }}'
+if ! curl -s -I $BASE_URL/coreos_production_pxe.vmlinuz | egrep -q '^HTTP/[0-9.]+ [23][0-9][0-9]' ; then
```
dghubble: swap to `grep -q -E '^HTTP/[0-9.]+ [23][0-9][0-9]'`

dghubble:
Can you fix the commit message according to https://github.com/coreos/coreos-baremetal/blob/master/CONTRIBUTING.md#format-of-the-commit-message?

```
scripts/get-coreos: Relax channel/version check

Your previous message is fine
```

And thanks!

PR #336, opened 2016-08-30
scripts/get-coreos: Make gpg binary customisable

urzds:
Distributions like Debian 8 ship a `gpg` (1.4.x) and a `gpg2` (2.1.x) binary,
 which both use the same config files, and thus cannot be used at the same
 time, due to incompatible options. Thus we allow the user to specify which
 gpg binary they want to use.


dghubble:
Merged as d56bf78e58220992508e08af1ccc2f24486edd7c

PR #340, opened 2016-09-07
examples/ignition: Fix etcd peer listen urls to use IPs

dghubble:
* See github.com/coreos/etcd/pull/6365

cc @gyuho


crawford:
Oof, binding on all interfaces isn't great. It's probably fine for the starter examples though. LGTM.

dghubble:
These are reference clusters, if you have a better suggestion can you please share? I've explored approaches (for other reasons) which introspect for the current default interface's IP which could be used as the binding address here, but folks haven't been happy with introspection. Alternatives make assumptions about other services on the network.

crawford:
> These are reference clusters

I think that is a fine stance to take. In order to get performance and security, the cluster needs to be tailored to its environment.

I think it's also reasonable to leverage coreos-metadata to determine exactly which interfaces should be bound, but that requires the user to modify the configs to match their environment.

dghubble:
Rebased and added etcd3 and torus too.

PR #346, opened 2016-09-16
Finished renaming Master -> controller

rothgar:
Examples are looking for k8s-master.yaml file. Also needed to change the profile ids for proper matching.


dghubble:
Thanks!

Its actually possible to eliminate the special k8s-controller-install Profile and use k8s-controller for either live or installed to disk clusters like I'm doing for rkt, but I'll test that out in a followup.

PR #348, opened 2016-09-19
scripts: Add devnet script to setup PXE/bootcfg bridge

dghubble:
The QEMU/KVM tutorial/development setup asks users to run `bootcfg` and `dnsmasq` (DHCP/TFTP/DNS) on the virtual bridge where QEMU/KVM VMs will be created. This script adds an optional "fast-track" which sets this up, but does it with proper rkt run transient units.

Run PXE services and bootcfg set to configure machines into an `etcd` example cluster.

```
$ ./scripts/devnet create etcd
Starting bootcfg with examples/groups/etcd mounted
Running as unit dev-bootcfg.service.
Starting dnsmasq to provide DHCP/TFTP/DNS services
Running as unit dev-dnsmasq.service.
```

Check logs as usual for systemd.

```
journalctl -u dev-bootcfg
journalctl -u dev-dnsmasq
```

Create the QEMU/KVM VMs with tutorial/development (known) hardware attributes.

```
./scripts/libvirt create
```

Stop and garbage collect the rkt run bootcfg and PXE services.

```
./scripts/devnet destroy
```

I believe this approach has advantages over some Procfile, Vagrantfile, or ISO ideas.

* Simple setup for developers using this daily
* Runs pods with rkt just like in production
* rkt run's within a systemd transient unit
    * Units are named `dev-bootcfg` and `dev-dnsmasq` for convenience
    * Journal logs directly with `journalctl -u dev-bootcfg`
    * More similar to production, where systemd/journald run rkt pods
* Solves the previous issue of always needing to run `rkt gc --grace-period=0`
    * Stop and rm the specific pod by UUID
    * Cleanup the transient system unit
* Avoids creating a whole VM just to rkt run these pods, the host has rkt
* Prepares the way for a few (optional) additional services I'll be adding soon

Though I hate to make the comparison, but you can think of this like a systemd and rkt integrated way of `docker-compose up` style management of a group of pods for bare-metal development.

### Followup

The tutorials/documentation are intentionally left as-is for now. Documentation clearly explains the purpose of each service and using `devnet` right off the bat may be too "magical". I'll work to strike a balance before updating docs so educational value is not lost.

### DIFF
--- a/scripts/devnet
+++ b/scripts/devnet
@@ -1,0 +1,100 @@
+#!/bin/bash
+# Create a virtual bridge with PXE services and bootcfg
+# USAGE: ./scripts/devnet create [example]
+# USAGE: ./scripts/devnet destroy
+set -u
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+EXAMPLE=${2:-}
+BRIDGE=metal0
+COREOS_VERSION=1153.0.0
+BOOTCFG_ARGS=""
+
+if [ "$EUID" -ne 0 ]
+  then echo "Please run as root"
+  exit
+fi
+
+function main {
+  case "$1" in
+    "create") create;;
+    "destroy") destroy;;
+    *)
+      usage
+      exit 2
+      ;;
+  esac
+}
+
+function usage {
+  echo "USAGE: ${0##*/} <command>"
+  echo "Commands:"
+  echo -e "\tcreate\tcreate bootcfg and PXE services on the bridge"
+  echo -e "\tdestroy\tdestroy the services on the bridge"
+}
+
+function check {
+  # SELinux, if present, it cannot be in Enforcing mode
+  if [ $(getenforce) == 'Enforcing' ]; then
+    echo "SELinux must be in permissive mode: 'setenforce Permissive'"
+    exit 1
+  fi
+
+  if [ ! -d $DIR/../examples/assets/coreos/$COREOS_VERSION ]; then
+    echo "Most examples use CoreOS Alpha $COREOS_VERSION. You may wish to download it with './scripts/get-coreos alpha $COREOS_VERSION'."
+  fi
+}
+
+function create {
+  check
+
+  if [ -z "$EXAMPLE" ]; then
+    echo "Starting bootcfg"
+  else
+    echo "Starting bootcfg configured to boot $EXAMPLE"
+  fi
+
+  if [ -z "$EXAMPLE" ]; then
+    # Mount a data volume with assets and enable gRPC
+    BOOTCFG_ARGS="-rpc-address=0.0.0.0:8081"
+    DATA_MOUNT="--volume data,kind=host,source=$(mktemp -d) \
+    --mount volume=assets,target=/var/lib/bootcfg/assets \
+    --volume assets,kind=host,source=$PWD/examples/assets,readOnly=true"
+  else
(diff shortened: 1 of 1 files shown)


dghubble:
@jonboulle thoughts?

I've noticed, `rkt stop` is not synchronous so sometimes the `rm` fails because the pod is running and another gc is needed.

```
rkt stop --uuid-file=/tmp/bootcfg --force
rkt rm --uuid-file=/tmp/bootcfg
```

scripts/devnet
```diff
@@ -0,0 +1,79 @@
...
+function destroy {
+  rkt stop --uuid-file=/tmp/bootcfg --force
+  rkt rm --uuid-file=/tmp/bootcfg
+  rkt stop --uuid-file=/tmp/dnsmasq --force
+  rkt rm --uuid-file=/tmp/dnsmasq
+  systemctl reset-failed dev-bootcfg
+  systemctl reset-failed dev-dnsmasq
+}
+
+main $@
```
joeatwork: Might want to add a trailing newline

scripts/devnet
```diff
@@ -0,0 +1,79 @@
...
+
+function check {
+  # SELinux, if present, it cannot be in Enforcing mode
+  if [ $(getenforce) == 'Enforcing' ]; then
+    echo "SELinux must be in permissive mode: 'setenforce Permissive'"
+    exit 1
+  fi
+
+  if [ ! -d $DIR/../examples/assets/coreos/$COREOS_VERSION ]; then
+    echo "Most examples use CoreOS $COREOS_VERSION. You may wish to download it."
```
joeatwork: COREOS_VERSION CHECK IS SUPER SWEET

scripts/devnet
dghubble: done

scripts/devnet
```diff
@@ -0,0 +1,79 @@
...
+    rkt run \
+    --uuid-file-save=/tmp/dnsmasq \
+    --net=metal0:IP=172.15.0.3 \
+    --mount volume=config,target=/etc/dnsmasq.conf \
+    --volume config,kind=host,source=$DIR/../contrib/dnsmasq/metal0.conf \
+    coreos.com/dnsmasq:v0.3.0
+}
+
+function destroy {
+  rkt stop --uuid-file=/tmp/bootcfg --force
```
joeatwork: Why the --force? (Question, not suggestion. I know like nothing about rkt)

joeatwork:
This startup totally works great on my machine, and is more convenient, less error prone, and more clear than the (pretty excellent) setup before! +1000!

scripts/devnet
dghubble: Good question, this may no longer be needed

dghubble:
Updated so that the example is optional,

    ./scripts/devnet create [EXAMPLE]

If no example is specific, bootcfg will be run with just an empty `/var/lib/bootcfg` and image assets, with the gRPC API enabled. This is useful for a vanilla, prod-like run.

scripts/devnet
dghubble: Without, if the stop hasn't finished it won't remove it. We just wanna remove it, we're only stopping it to be nice.

```
"2516b402-db39-45ea-bb95-dab8fe196822"
rm: pod "2516b402-db39-45ea-bb95-dab8fe196822" is currently running
rm: failed to remove one or more pods
```

joeatwork: Sounds good to me!

PR #354, opened 2016-09-26
[doc] Fix sample Docker command

ebraminio:
Highly minor issue, but better to take more time from anyone else, I guess


dghubble:
Thanks

PR #362, opened 2016-10-15
Includable metadata bundles

### LINKED ISSUE #298: Add representation for bundles of machine metadata
Certain sets of machine metadata are often used together. It would be useful to refer to those by name to facilitate re-use, while still keeping machine profiles fairly generic. For example, a CoreOS release could be represented this way.

Bundle of metadata in /var/lib/bootcfg/metadata/coreos1053.json

```
{
  "id": "coreos1053",
  "name": "coreos1053",
  "metadata": {
    "kernel": "kernel_uri",
    "kernel_sig": "kernel_sig_uri",
    "initrd": "initrd_uri",
    "initrd_sig": "initrd_sig_uri",
    "image":  "image_uri",
    "image_sig":  "image_sig_uri"
  }
}
```

Possible Usage:

```
{
  "id": "node1",
  "name": "etcd Node 1",
  "profile": "etcd",
  "selector": {
    "mac": "52:54:00:a1:9c:ae"
  },
  "metadata": {
    "etcd_name": "node1",
    "etcd_initial_cluster": "node1=http://172.15.0.21:2380,node2=http://172.15.0.22:2380,node3=http://172.15.0.23:2380"
  },
  "metadata_include": [
    "coreos1053"
  ]
}
```

rel. #286

nickethier:
This attempts to resolve #298 

This allows separate metadata to be defined under `<data-path>/metadata/<id>.json` and subsequently included in group definitions by using the `metadata_include` key and referencing metadata files by `<id>`.

Includes some docs and tests.


nickethier:
Rebased against master, /cc @dghubble

dghubble:
Thanks for your submission. Looks like a good attempt to implement the described idea in #298. We're not ready to commit to that api decision or prioritize time on that right this moment though. We're working on coreos-baremetal shipping as part of Tectonic Enterprise and I'd like to see the templating fully built out and merges and lists before committing on this one. This may simmer for a bit, but we should return to this later.

nickethier:
Sounds good, I may keep this merged into master on my fork and maintain it there for now as it solves my needs pretty well. I had planned on implementing support for merging lists and maps so I'll be sure to push it to this branch if its something you folks could use later down the road. Thanks for the great software!

dghubble:
I don't think its likely we'll build more any templating features into Matchbox itself. Client tools like Terraform or others handle this detail, leaving Matchbox to just authenticate clients, hold configs, and serve them to PXE booting machines.

Basically, letting this idea simmer, I don't think its the direction to go. Every app wants to be the templating layer. Matchbox won't remove its templating features, but isn't going to push and develop that use case imo.

Thanks so much for the effort. I'd be interested to hear opinions.

PR #365, opened 2016-10-19
Documentation: Show self-hosted Kubernetes upgrade process

dghubble:
I've upgraded several clusters to v1.4.3 and documented that process here.

cc @aaronlevy @sym3tri

### DIFF
--- a/Documentation/bootkube-upgrades.md
+++ b/Documentation/bootkube-upgrades.md
@@ -1,0 +1,175 @@
+
+# Upgrading Self-hosted Kubernetes
+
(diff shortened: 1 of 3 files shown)


sym3tri:
Why is this in the bare-metal repo? Seems relevant to more than just bare-metal.

aaronlevy:
Yeah, this would be great to have in bootkube repo too

dghubble:
Yeah, it should work the same on cloud providers. I've performed the migration on QEMU/KVM clusters used in the self-hosted tutorial in this project and on real bare-metal clusters. This shouldn't matter for upgrades, but both those clusters still use https://github.com/kubernetes-incubator/bootkube/pull/151

dghubble:
Do you want it, but without the bare-metal mentions? I can send a PR

sym3tri:
maybe we put on coreos.com or tectonic.com?

dghubble:
https://github.com/kubernetes-incubator/bootkube/pull/159

PR #370, opened 2016-11-02
examples: Use Kubelet --pod-manifest-path and EnvironmentFile

dghubble:
* Use the Kubelet `--pod-manifest-path`, `--config` is deprecated in Kubernetes 1.4+
* Self-hosted Kubernetes clusters expect the Kubelet ACI and VERSION to be read from an environment file on disk


dghubble:
cc @derekparker

derekparker:
LGTM

PR #375, opened 2016-11-17
get-coreos: Resume partial downloads if interrupted.

jpwarren:
If the get-coreos script is interrupted, partial downloads are ignored when you run it again and the entire process starts from scratch, which downloads ~500 MB of data all over again.

This pull-request adds the -C - flags to the curl commandline so that curl will resume partially completed downloads, saving bandwidth and time.


dghubble:
```
CoreOS Image Signing Key
######################################################################## 100.0%
curl: (33) HTTP server doesn't seem to support byte ranges. Cannot resume.
```

and if we ignore that, a sig can be mangled as well

```
coreos_production_image.bin.bz2.sig
######################################################################## 100.0%
gpg: [don't know]: invalid packet (ctb=3c)
gpg: assuming signed data in `/home/dghubble/workspace/go/src/github.com/coreos/coreos-baremetal/examples/assets/coreos/1192.2.0/coreos_production_pxe.vmlinuz'
gpg: Signature made Thu 20 Oct 2016 07:08:14 PM PDT
gpg:                using RSA key 48F9B96A2E16137F
gpg: BAD signature from "CoreOS Buildbot (Offical Builds) <buildbot@coreos.com>"
```

jpwarren:
Huh. Weird that it worked at all when I tried it.

I can see how keeping the fetch of the signature sortof atomic is reasonable, and they're relatively tiny.

We're not all on gigabit fibre, so having to download large images with no resume is suboptimal. This seems to be an infrastructure problem at coreos.com. Do we need to chat to whomever maintains the webservers to enable byte range support?

Or is there a better solution?

jpwarren:
Potential script change would be to detect byte range support and use it if available, and if not, then don't attempt.

dghubble:
I'll ask someone to look into resumption from the site. The images are served from buckets.

dghubble:
Apparently this may be related to their being served through cloudfront. Perhaps skip resumption for sigs

dghubble:
curl resume and range request implementation handling of completely downloaded files seems unreliable. Some curl issues mention it. I see this problem downloading other distros images as well.

```
$ curl -# -C - https://stable.release.core-os.net/amd64-usr/current/coreos_production_image.bin.bz2.sig -o target
cp target original
$ curl -# -C - https://stable.release.core-os.net/amd64-usr/current/coreos_production_image.bin.bz2.sig -o target
$ cmp original target 
cmp: EOF on original
543 Dec  4 14:30 original
760 Dec  4 14:30 target
```

I even have a fairly recent curl too.

```
curl 7.47.1 (x86_64-redhat-linux-gnu) libcurl/7.47.1 NSS/3.27 zlib/1.2.8 libidn/1.33 libpsl/0.13.0 (+libidn/1.33) libssh2/1.7.0 nghttp2/1.7.1
```

In this case, GPG correctly complains that the signature becomes invalid.

For now, I prefer to leave the script without continue-at. Datacenters have bandwidth to download an iso into their cache. Admins shouldn't spend time debugging mangled images or have to question the signature legitimacy. Plenty of those saints are on ancient boxes with ancient curl already.

cc @crawford

PR #381, opened 2016-11-21
examples/ignition: check response status code during curl invocation

nailgun:
Currenty `curl` return with successful exit code even if server returns non-successful response, such as 404 or 500. I found this when `k8s-certs@` services succeeded, however I didn't prepared any credentials.

`--fail` option checks response status code and modifies curl exit code


dghubble:
Solved in #470, thanks!

PR #382, opened 2016-11-21
fix scripts/tls/kube-conf for darwin OS type

nailgun:
On my macOS 10.12.1 laptop `OSTYPE=darwin16`. This fix should handle all darwin OSes.


dghubble:
Thanks, merged in 2bdffc7569a3fea62696ddc991e1d619a64361cd

PR #384, opened 2016-11-23
Update pixiecore link

Quentin-M:


dghubble:
Fixed in a78c3a0f7562edabf2bc3ad9f8e8fc3e2e825937, though iPXE/GRUB are definitely the focus.

PR #387, opened 2016-12-04
Documentation: update deployment.md cert-gen references

bzub:
Deployment document refers to `scripts/tls` for self-signed certificates.
The current location is actually `examples/etc/bootcfg`


dghubble:
Thanks

PR #395, opened 2016-12-09
examples: Remove old kubelet pods by uuid-file

dghubble:
* Save the rkt pod uuid on start and remove pod resources (files, network) on restart, without waiting on gc


dghubble:
Actually fixed in eb9809ee866d4737981f9fe8d1816ebaf7af70b3

PR #397, opened 2016-12-12
Documentation: Deprecate Pixiecore support

dghubble:
* coreos-baremetal must support production Tectonic customers on bare-metal
* Focus on real-world, varied network environments and flexibility
* iPXE provides better hardware introspection than being limited to MAC
* Matching must extend beyond hardware identifiers
* ISC DHCP and dnsmasq provides production DHCP service
* If we desire all-in-one ease, it will be accomplished by packaging
* Drop Pixiecore (all-in-one binary) from the setups we can support or recommend
    * Code remains so anyone who was using it isn't taken by surprise and there are no hard feelings :)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Latest
 
+* Deprecate Pixiecore support
 * Update Fuze and Ignition to v0.11.2
 
 ## v0.4.2 (2016-12-7)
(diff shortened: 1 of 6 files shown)


dghubble:
Got a question about this deprecation so I'll provide a bit more detail:

matchbox pixiecore support was neat and I don't want to knock the project or their efforts. However, its unheard of in datacenter environments, it re-implements DHCP/TFTP when quality implementations already exist, it doesn't make use of iPXE, it has oddities for use with matchbox (only being able to use MAC's), its statefulness conflicts with how we'd like to inventory machines long term, and ultimately it isn't something we'd like to support going forward.

I believe a better long term approach for machine stats is a machine inventory service which is independent of both PXE network services and provisioning services.

PR #402, opened 2016-12-15
Minor doc improvement

dol:
Exit `openssl s_client` with the help of echo. Without the stdin from the echo command the openssl command hangs and needs to be quit manually.


dghubble:
I'm not sure I follow. If I mangle hosts, ports, or expected files the client complains loudly. In what cases is this hanging for you?

PR #404, opened 2016-12-18
(maybe) examples: Use the kube-addon-manager on static clusters

dghubble:
* Use kube-addon-manager to create Kubernetes addons on static Kubernetes clusters, rather than a systemd unit and curl script

Background:

The on-host kubelet, configured with `--pod-manifest-path=/etc/kubernetes/manifests`, creates and syncs pod manifests onto the *host* node. The `kube-addon-manager` pod ensures `/etc/kubernetes/addons` manifests are created on the *cluster* (such as kube-dns and heapster). The `kube-addon-manager` can be added to the controller pod-manifest-path. One will run on each controller (though one per cluster would suffice).

Pros:

The `kube-addon-manager` is a bit better than the curl scripts it replaces. It periodically syncs and retries and uses `kubectl` to create different resources. It elevates manifest creation failures from a journal log on the host to something visible in kubectl logs.

Cons:

However, it does add yet another gcr.io image to fetch (EDIT: that image is huge, 255MB are you kidding me!), and its not much more than a shell script itself.

Note: This is only even plausible on static Kubernetes clusters. In self-hosted clusters, the idea is to break away from pod manifests on the host and kube-addon-manager is not a good fit.

@pbx0 thoughts? I'm thinking maybe we fork the addon-manager and make it tiny or leave the curl scripts we had.

### DIFF
--- a/examples/ignition/k8s-controller.yaml
+++ b/examples/ignition/k8s-controller.yaml
@@ -100,16 +100,6 @@ systemd:
         RestartSec=10
         [Install]
         WantedBy=multi-user.target
-    - name: k8s-addons.service
-      enable: true
-      contents: |
-        [Unit]
-        Description=Kubernetes Addons
-        [Service]
-        Type=oneshot
-        ExecStart=/opt/k8s-addons
-        [Install]
-        WantedBy=multi-user.target
     {{ if eq .container_runtime "rkt" }}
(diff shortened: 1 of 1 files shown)


dghubble:
For now, I'm going to close this. I'm not entirely happy with kube-addon.

PR #407, opened 2016-12-30
examples: Upgrade self-hosted Kubernetes to v1.5.1

dghubble:
* Update bootkube version to stop requiring the forked binary; bootkube now allows DNS names
* https://github.com/kubernetes-incubator/bootkube/pull/151

cc @aaronlevy

### DIFF
--- a/Documentation/bootkube.md
+++ b/Documentation/bootkube.md
@@ -1,7 +1,7 @@
 
 # Self-Hosted Kubernetes
 
-The self-hosted Kubernetes example provisions a 3 node "self-hosted" Kubernetes v1.4.7 cluster. On-host kubelets wait for an apiserver to become reachable, then yield to kubelet pods scheduled via daemonset. [bootkube](https://github.com/kubernetes-incubator/bootkube) is run on any controller to bootstrap a temporary apiserver which schedules control plane components as pods before exiting. An etcd cluster backs Kubernetes and coordinates CoreOS auto-updates (enabled for disk installs).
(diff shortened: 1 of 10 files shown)


examples/ignition/bootkube-controller.yaml
```diff
@@ -133,7 +133,7 @@ storage:
           # Wrapper for bootkube start
           set -e
           BOOTKUBE_ACI="${BOOTKUBE_ACI:-quay.io/coreos/bootkube}"
-          BOOTKUBE_VERSION="${BOOTKUBE_VERSION:-v0.2.6}"
+          BOOTKUBE_VERSION="${BOOTKUBE_VERSION:-v0.3.0}"
```
aaronlevy: Maybe minor, but wanted to note that you're using two different versions of bootkube (based on the referenced hash in the docs). Assuming this is fine because the docs reference a commit that is not in a release yet, but needed for the `render` step -- whereas the container version here is only used for the `start` step -- which is unchanged. In this case seems fine, but in future should try and sync these up to avoid hard to debug / unknown states.

aaronlevy:
One note about version skew, but otherwise lgtm

examples/ignition/bootkube-controller.yaml
dghubble: Talked IRL. If we could make a4739ed48f09ad75309479660bbe0ef7778c266c into a v0.3.1, those could be aligned and it would also simplify the Tectonic bump to v0.3.1.

dghubble:
Updated to bootkube v0.3.1 and ran tests again.

Documentation/bootkube.md
```diff
@@ -12,10 +12,10 @@ Ensure that you've gone through the [bootcfg with rkt](getting-started-rkt.md) o
 * Create the example libvirt client VMs
 * `/etc/hosts` entries for `node[1-3].example.com` (or pass custom names to `k8s-certgen`)
 
-Build and install the [fork of bootkube](https://github.com/dghubble/bootkube), which supports DNS names.
+Build and install [bootkube](https://github.com/kubernetes-incubator/bootkube).
```
aaronlevy: You could also link to the binary release tarball

PR #408, opened 2017-01-05
scripts: Add virt-install flag so reboots work

dghubble:
* virt-install created VMs which reboot themselves should actually reboot in order to mirror real machines

cc @ericchiang @rithujohn191 this should improve our lives


rithujohn191:
lgtm :+1:

Issue #92, opened 2016-03-08
Template network boot configurations (iPXE, GRUB)


dghubble:
Allow a grub config or iPXE script file with template elements to be provided directly. Substitute template values from `selector` and/or `metadata`. Currently, a `NetBoot` struct and internal templates determine the iPXE scripts and GRUB configs which are rendered which isn't as customizable.

dghubble:
This is needed for the Kubernetes profiles to be practical via Grub netboot for UEFI. The work-around for now is to write different profiles for each machine which embeds a MAC or UUID in the `coreos.config.url` kernel argument.

dghubble:
Prioritizing iPXE usage right now and improving GRUB use cases next.

ropsoft:
Cool. I was about to open an issue asking if this exists currently! Two questions if I may...:

1.) To make sure I'm not mistaken on the goal here... would this allow for setting metadata to env vars?

```
{
  "id": "coreos-install",
  "name": "CoreOS Install",
  "profile": "install-reboot",
  "metadata": {
    "my_custom_var": "${custom_env_var}"
    "coreos_channel": "alpha",
    "coreos_version": "1109.1.0",
    "ignition_endpoint": "http://bootcfg.foo:8080/ignition",
    "baseurl": "http://bootcfg.foo:8080/assets/coreos"
  }
}
```

2.) Would the env var be evaluated every time the group is matched or just templated in memory to static text a single time when bootcfg starts?

dghubble:
Hm, you mean populating metadata from environment variables from the bootcfg process environment? No, this is more for being able to write an iPXE config or GRUB config yourself and have it be templated with metadata variables, much like the Ignition, Cloud-config, and generic configs today. 

The iPXE and GRUB configs are created from a hardcoded template https://github.com/coreos/coreos-baremetal/blob/master/bootcfg/http/ipxe.go#L17. This has been ok in most cases, but it might be useful to customize that. The "boot" section in a profile is a bit of a special beast, when we might have an advanced mode where a profile can refer to a named iPXE config resource. https://github.com/coreos/coreos-baremetal/blob/master/Documentation/bootcfg.md#profiles

SchwarzM:
it would be nice to see this we could need a hook in this to do additional ipxe scripting before boot.

discostur:
Is there any progress in here? If i try to boot via iPXE i'm getting the following error on the iPXE-Client:

```
Error 0x040ee119
```

[...]

PR #410, opened 2017-01-05
ipxe: use `mac` instead of `net0/mac`

mcuadros:
All the ipxe, scripts and examples are using `net0/mac` as mac selector variable, when you have several interfaces and net0 is not the interface being use, the mac sent as selector, could be useless/unknown as selector.

Using just `mac` variable, the mac of the interface being use is returned, solving the issue.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -11,7 +11,7 @@ Serves a static iPXE boot script which gathers client machine attributes and cha
 **Response**
 
     #!ipxe
-    chain ipxe?uuid=${uuid}&mac=${net0/mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial}
+    chain ipxe?uuid=${uuid}&mac=${mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial}
 
 Client's booted with the `/ipxe.boot` endpoint will introspect and make a request to `/ipxe` with the `uuid`, `mac`, `hostname`, and `serial` value as query arguments.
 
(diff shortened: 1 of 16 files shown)


dghubble:
Yeah, this seems like a better default. Ours align by convention, but its just a convention. We'll test this some more.

dghubble:
This probably requires more consideration, I'll open an issue. OpenStack learned lessons in the past we could apply here. https://bugs.launchpad.net/ironic/+bug/1504482

mcuadros:
Another options is use `netX` but is only available after the boot/dhcp event, if you run `config` before boot, you can see clearly that is not available. In my case this works in the same way that work `mac`

These is how looks my boot script:
```
#!ipxe
set vlan 2001

set idx:int32 0
:loop isset ${net${idx}/mac} || goto loop_done
echo Configuring VLAN ${vlan} on interface net${idx}
vcreate --tag ${vlan} net${idx}
dhcp net${idx}-${vlan} && goto boot || inc idx && goto loop
:loop_done
goto failed

:boot
echo Booting at interface ${netX/ifname}
set server http://${next-server}:8080/
chain ${server}ipxe?uuid=${uuid}&mac=${mac:hexhyp}&domain=${domain}&hostname=${hostname}&serial=${serial} || goto failed

:failed
echo ERROR: Unable to boot at any interface
prompt --key 0x02 --timeout 2000 Press Ctrl-B for the iPXE command line... && shell ||
```

This remembers me that could be interesting be able to customize the ipxe script being server

dghubble:
Admittedly, the `/boot.ipxe` endpoint is minimal and optional - it just provides a pretty [endpoint](https://github.com/coreos/coreos-baremetal/blob/master/Documentation/network-setup.md#ipxe) for chaining to the `/ipxe` config endpoint. I'd expect lots of folks are chainloading directly in their own way like you are and that seems perfect.

Yep, its desired to let the ipxe and grub configs be templated and customized per-machine. #92

dghubble:
Rebased and merged as cedeb868f9be0fbd15376c9f6c2a637b85fb97d5, thanks

dghubble:
Rebased and merged as cedeb868f9be0fbd15376c9f6c2a637b85fb97d5, thanks!

PR #411, opened 2017-01-05
examples: Update Kubernetes to use self-hosted flannel

dghubble:
* Self-hosted Kubernetes now runs flannel as a daemonset

This may require a return to the dghubble fork of bootkube for render, to remove the api-server advertise address, but ideally not.

cc @aaronlevy @pbx0

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@ Rename `bootcfg` to CoreOS `matchbox`!
 
 * Upgrade Kubernetes v1.5.1 (static) example clusters
 * Upgrade Kubernetes v1.5.1 (self-hosted) example cluster
+* Switch Kubernetes (self-hosted) to run flannel as pods
 * Combine rktnetes Ignition into Kubernetes static cluster
 
 #### Migration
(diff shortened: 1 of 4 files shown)


dghubble:
Don't update anything until https://github.com/kubernetes-incubator/bootkube/issues/262 is resolved.

Quentin-M:
@dghubble FWIW I believe the issue you mentioned existed before the self-hosted changes and has been introduced by the recent change in the pod checkpointer.

Quentin-M:
i.e. generic pod checkpoint impl. https://github.com/kubernetes-incubator/bootkube/pull/252 https://github.com/kubernetes-incubator/bootkube/pull/238 etc.

ericchiang:
Bumping to bootkube v0.3.3 seems to have worked. It survived a reboot at least with the libvirt setup. Going to bug aaron about more tests we can run.

aaronlevy:
The issue in v0.3.3 is that you can't delete last apiserver pod: https://github.com/kubernetes-incubator/bootkube/issues/267

ericchiang:
Updating to v0.3.4 and testing.

ericchiang:
This works but bootkube hasn't done a binary release yet. Since the docs suggest wget'ing the release tarball, we should wait to merge until that's in.

aaronlevy:
Forgot to attach the tarball -- added to release.

dghubble:
Thanks. Verified the apiserver comes back after controller reboots and after explicitly deleting the api-server so those regressions do seem to be solved. The switch to self-hosted flannel does increase the time it takes for the cluster to come up and some crash looping is expected, but resolves.

LGTM

dghubble:
c'mon travis...

PR #422, opened 2017-01-17
Switch to a Makefile driven develop/release process

dghubble:
* Add make targets for vendor, docker-image, and tools
* Move scripts into the scripts folder


Makefile
```diff
@@ -1,32 +1,67 @@
...
+IMAGE_REPO=coreos/matchbox
+QUAY_REPO=quay.io/coreos/matchbox
 
-codegen: tools
-	./scripts/codegen
+all: build
+
+build: clean bin/matchbox bin/bootcmd
 
 bin/matchbox:
```
crawford: These two can be collapsed into a target pattern.

Makefile
```diff
@@ -1,32 +1,67 @@
...
-	go build -o bin/matchbox -ldflags $(LD_FLAGS) -a github.com/coreos/matchbox/cmd/matchbox
+	@go build -o bin/matchbox -v -ldflags $(LD_FLAGS) $(REPO)/cmd/matchbox
```
crawford: You might want to use a variable instead of `@`. That way, you can turn on or off verbose building. Check out https://github.com/crawford/arm-toolkit/blob/be1984fabddaeff4777ff96b7f6a5830ee760102/makefile.mk#L40 as an example.

Makefile
```diff
@@ -1,32 +1,67 @@
...
+tools: bin/protoc bin/protoc-gen-go
+
+bin/protoc:
+	@./scripts/get-protoc
+
+bin/protoc-gen-go:
+	@go build -o bin/protoc-gen-go $(REPO)/vendor/github.com/golang/protobuf/protoc-gen-go
+
+clean:
+	@rm -rf bin
```
crawford: I prefer to explicitly remove each file from the build directory and then remove the directory only if it is empty. This only works if make knows about every generated file, so this might not work here.

Makefile
```diff
@@ -35,30 +70,30 @@ release: \
 # matchbox
 
 bin/linux-amd64/matchbox:
```
crawford: Target pattern

Makefile
dghubble: sure

Makefile
dghubble: Both the binary name and architecture are used in the recipe. It could probably be parsed out, but for now this release process is unchanged and fairly readable.

Makefile
dghubble: Hm, we could list all these out:

```
	@rm -f bin/matchbox
	@rm -f bin/bootcmd
	@rm -f bin/protoc
	@rm -f bin/protoc-gen-go
	@rmdir bin
```

Also, the target still needs to succeed if bin doesn't exist too.

Makefile
crawford: You can do something like this:

```
bin/linux-amd64/matchbox: OS=linux ARCH=amd64
bin/linux-arm/matchbox: OS=linux ARCH=arm
...

bin/%/matchbox:
        GOOS=$(OS) GOARCH=$(ARCH) go build -o $@ -ldflags $(LD_FLAGS) -a $(REPO)/cmd/matchbox
```

dghubble: I'm not sure the 2nd env variable "dependency" gets picked up this way

crawford: I forgot that you could only have one per target. You'd need to write this:

```make
bin/linux-amd64/matchbox: OS=linux
bin/linux-amd64/matchbox: ARCH=amd64
bin/linux-arm/matchbox: OS=linux
bin/linux-arm/matchbox: ARCH=arm
...

bin/%/matchbox:
        GOOS=$(OS) GOARCH=$(ARCH) go build -o $@ -ldflags $(LD_FLAGS) -a $(REPO)/cmd/matchbox
```

Or you could combine all of these into a single variable:

```make
bin/linux-amd64/matchbox: GOARGS = GOOS=linux GOARCH=amd64
bin/linux-arm/matchbox: GOARGS = GOOS=linux GOARCH=arm
...

bin/%/matchbox:
        $(GOARGS) go build -o $@ -ldflags $(LD_FLAGS) -a $(REPO)/cmd/matchbox
```

dghubble: But it was kinda clear before right? I'm for Makefiles, but some of the features one might use are so opaque

dghubble: GOARGS might be alright I guess

Makefile
dghubble: I mean, I see your condition, find, xargs, exit 0.

Makefile
dghubble: added

Makefile
dghubble: neat'o

dghubble: neato, but I think I'll pass if thats fine

dghubble:
Objections?

PR #427, opened 2017-01-19
Lock architecture to ARMv6 for linux-arm images

urzds:
Unless `GOARM` is specified, the Go compiler will choose the compiling system's architecture as a target (similar to GCC's `-march=native`). This commit prevents that by explicitly selecting ARMv6 (e.g. Raspberry Pi 1) as the target.

According to the documentation, ARM64 does not need this, as ARMv8 is implicit and the value of `GOARM` is not considered for this architecture.

See-Also: https://github.com/golang/go/wiki/GoArm


dghubble:
LGTM

PR #433, opened 2017-02-07
Update the grub config.

cjyar:
I'm not sure if the existing EFI Grub config should just be deleted. But this will try EFI and then fall back to old-fashioned BIOS. Tested on my non-EFI setup.


dghubble:
Thanks very much for this and apologies for the delay. Yes, we should use the Profile's "args" field in GRUB configs since it replaces the "cmdline" field. I've rebased your change and it should go in v0.6.0 soon.

dghubble:
Merged as 2a83612ffbf3af80314fc381b74162657ed837b8

PR #440, opened 2017-02-17
Switch from etcd2 to etcd3

dghubble:
Changes to `examples/*`:

* Update `etcd3` and `etcd3-install` example clusters
   * Use the etcd-member unit which ships on CoreOS (1214.0.0+)
   * Switch from using etcd proxy to using etcd3 gateway
* Remove `etcd` and `etcd-install` example clusters (i.e. etcd2)
   * This project doesn't aim to maintain etcd2 examples going forward
* Update `bootkube` and `bootkube-install` examples to use bootkube `v0.3.7` which uses etcd3

*note: Only self-hosted Kubernetes clusters are updated to etcd3, not static kubernetes or rktnetes.*
@gyuho @pbx0


gyuho:
`etcd3` part looks good to me. /cc @xiang90

PR #453, opened 2017-03-08
cmd/matchbox: fix -address parsing when built with Go 1.8

dghubble:
* 1.8 changed the behavior of url.Parse so it is no longer appropriate for parsing values like 0.0.0.0:8080
* Pass the address directly to http.ListenAndServe which gives reasonable errors when bad values are used


dghubble:
Go 1.8 address parsing fixed in #453.

PR #454, opened 2017-03-08
Documentation/deployment.md: Added note for tectonic based matchbox example

soggiest:
The Tectonic matchbox service example relies on rkt being installed on the local host otherwise the matchbox service will fail. Added a note to the Tectonic matchbox service example section and provided a link to the rkt installation documentation.

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -113,6 +113,8 @@ Environment="MATCHBOX_RPC_ADDRESS=0.0.0.0:8081"
 
 The Tectonic [Installer](https://tectonic.com/enterprise/docs/latest/install/bare-metal/index.html) uses this API. Tectonic users with a CoreOS provisioner can start with an example that enables it.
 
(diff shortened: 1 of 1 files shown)


dghubble:
That section is about the matchbox `-grpc-address` and gRPC API feature (which happens to be used by Tectonic). The matchbox gPRC API can be used whether matchbox is run as a binary, docker image, rkt image, etc.

dghubble:
How about:

```
Tectonic users with a CoreOS provisioner may use the following `matchbox-for-tectonic.service` example:
```

Also, can you squash into one commit and follow the commit message style guide? Thanks

soggiest:
Do we want to exclude RPM based installations from using the matchbox tectonic example?

dghubble:
I think "Tectonic users with a CoreOS provisioner" makes it clear the convenience example is for CoreOS. There is a separate section for RPM-based distros.

I'd prefer the official Tectonic [docs](https://coreos.com/tectonic/docs/latest/install/bare-metal/index.html) be the official guide for Tectonic use cases though so feel free to drop in a call to action and link instead.

dghubble:
I think the current docs look alright so I'm closing. Feel free to open a PR if you have changes.

Issue #297, opened 2016-08-03
Document a cache proxy setup for images


dghubble:
Production users may outgrow bootcfg's assets folder feature, which is primarily for small labs and development image serving. A recommended approach is to use a cache proxy container which serves and caches requested images and can be allowed to egress the network in a controlled way.

Add a container image, docs, and usage guide to contrib.

rel. #286

dghubble:
A work in progress from @kbrwn https://github.com/kbrwn/rkt-squid

danehans:
I would like to use a squid proxy to host my bootcfg coreos images instead of the local assets dir. It looks like https://github.com/kbrwn/rkt-squid has been inactive for a bit. I am looking at https://github.com/sameersbn/docker-squid as another option.

I am trying to understand the workflow. Would I run the get-coreos script from my squid proxy to cache the kernel/initrd, verify sigs, etc.. and simply update the initrd/kernel values in my profiles to use the actual coreos urls:

**profile example:**
`{
  "id": "bootkube-controller",
  "name": "bootkube Ready Controller",
  "boot": {
    "kernel": "https://beta.release.core-os.net/amd64-usr/1185.1.0/coreos_production_pxe.vmlinuz",
    "initrd": ["https://beta.release.core-os.net/amd64-usr/1185.1.0/coreos_production_pxe.vmlinuz"],
...
    }
  },
  "cloud_id": "",
  "ignition_id": "bootkube-controller.yaml"
}`

cc @kbrwn

dghubble:
I don't think you need to load the files with `get-coreos` if you're running it as an HTTP cache proxy. Anywhere those images are pulled will need to respect the proxy though.

danehans:
@dghubble I'm trying to figure out how to pass the proxy setting to the server being managed by bootcfg. Do I simply add something like the following to the group metadata:

`  "metadata": {
    "HTTP_PROXY": "http://proxy.example.com:3128",
    ..
  }`

dghubble:
Sure, then be sure to choose what to do with that metadata value during machine provisioning. Like setting the `http_proxy`, `https_proxy`, or `all_proxy` environment variables for apps. CoreUpdate, rkt, and Docker should respect it.

https://coreos.com/os/docs/latest/update-strategies.html#updating-behind-a-proxy
https://coreos.com/os/docs/latest/customizing-docker.html#use-an-http-proxy

[...]

PR #455, opened 2017-03-08
Adds documentation for using a Squid proxy with Matchbox.

danehans:
This PR closes https://github.com/coreos/matchbox/issues/297

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -233,7 +233,7 @@ $ curl http://matchbox.example.com:8080/assets/coreos/1235.9.0/
 <pre>...
 ```
 
-For large production environments, use a cache proxy or mirror suitable for your environment to serve CoreOS images.
+For large production environments, use a cache proxy or mirror suitable for your environment to serve CoreOS images. Review the cache proxy [guide](https://github.com/coreos/matchbox/blob/master/Documentation/proxy-setup.md) for details.
 
 ## Network
 
--- a/Documentation/proxy-setup.md
+++ b/Documentation/proxy-setup.md
@@ -1,0 +1,97 @@
+# Cache Proxy
+
+This guide shows how to setup a [Squid](http://www.squid-cache.org/) cache proxy for providing kernel/initrd files to PXE, iPXE, or GRUB2 client machines. This setup runs Squid as a Docker container using the [sameersbn/squid](https://quay.io/repository/sameersbn/squid)
+image.
+
+The Squid container requires a squid.conf file to run. Download the example squid.conf file from the [sameersbn/docker-squid](https://github.com/sameersbn/docker-squid) repo:
+```
+curl -O https://raw.githubusercontent.com/sameersbn/docker-squid/master/squid.conf
+```
+
+Squid [interception caching](http://wiki.squid-cache.org/SquidFaq/InterceptionProxy#Concepts_of_Interception_Caching) is required for proxying PXE, iPXE, or GRUB2 client machines. Set the intercept mode in squid.conf:
+```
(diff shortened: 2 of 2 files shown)


ndouglas:
I think I'll need to go back through and test this in greater depth, perhaps later today or tomorrow.

Briefly:
Your instructions are very clear and explain the setup very well.

I'm running Squid with a manual command (basically copied and pasted from your service definition, but with the path to squid.conf updated appropriately) on a clean + docker + docker-compose installation of Debian Jessie in a pretty simple VM running on ESXi 6.5.  The iptables rules were applied.  My DHCP is provided by pfSense, and my group definition was updated and the client boots and is pointed to the correct URL.  It times out, though, on retrieving the kernel.

The issue seems to be with the Squid configuration -- it doesn't seem to be actively listening or opening the port.  Telnetting to either :80 or :3128 from the local machine fails.  I see a repeated error when running the container in the foreground:

`2017/03/09 23:04:28| ERROR: No forward-proxy ports configured.`

This is probably PEBKAC on my part, but I wanted to mention it before doing more in-depth (and careful!) testing tomorrow, just in case it rang any bells.  I'm not very knowledgeable about Squid, so if this points clearly to an obvious issue, that's why I'm missing it.  I might just need to add a port or something to the config.

I also have CentOS and CoreOS VMs handy, and I can whip anything else up pretty easily if you have any suggestions.

danehans:
@ndouglas make sure you have this line in your squid.conf file:

`http_port 3128 intercept`

In my example, I am using the `--net=host` flag so the squid container uses the host's netns. I can netstat the port and see it used by the squid container:

`$ sudo netstat -tulpn | grep 3128
tcp        0      0 0.0.0.0:3128            0.0.0.0:*               LISTEN      6482/squid3  `

Verify your firewalld/iptables settings. My setup uses firewalld:
```
$ sudo firewall-cmd --get-active-zones
internal
  interfaces: enp14s0
public
  interfaces: enp15s0

$ sudo firewall-cmd --zone=internal --list-all
internal (active)
  target: default
  icmp-block-inversion: no
  interfaces: enp14s0
  sources: 
  services: dhcp dhcpv6-client dns ipp-client mdns samba-client ssh tftp
  ports: 3128/tcp 4080/tcp 8888/tcp 67/udp 53/udp 3129/tcp 8080-8082/tcp
  protocols: 
  masquerade: no
  forward-ports: port=80:proto=tcp:toport=3128:toaddr=10.30.118.164
	port=80:proto=tcp:toport=3128:toaddr=10.10.129.133
  sourceports: 
  icmp-blocks: 
  rich rules: 
```

Although I opened 3129 in the example above, I am only using TCP port 3128 for squid. You may also want to try running standard squid proxy on port 3128 and transparent proxy on 3129. Here is the squid.conf settings:
```
http_port 3128
http_port 3129 transparent
```

Good luck and let me know how it goes.

dghubble:
Thanks! Looks great, I'll try to find some time to vet this and use it for real and add onto it if that's alright.

I'd like a usage section to show rkt and docker one-liners and to fold the configs and systemd units into a contrib directory and expand on how networked nodes or individual nodes are configured to use the proxy. Also, these docs now get published to [coreos.com](https://coreos.com/matchbox/docs/latest) so I'll figure out where they fit in.

danehans:
Sounds good. I'll try to do another round of vetting myself.

dghubble:
I haven't had the time I had hoped. Let's merge these as drafts in `contrib/squid` (see #510) so we can use #511 to track iterative improvements toward being sync'd to the docs site.

PR #457, opened 2017-03-10
examples: Update self-hosted Kubernetes to v1.5.4

dghubble:
* Upgrade to self-hosted Kubernetes v1.5.4
* Use bootkube v0.3.10
* Disable anonymous-auth flag for on-host kubelet
* Set the client CA for the on-host kubelet, based on the kubeconfig

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@ Notable changes between releases.
 ### Examples
 
 * Upgrade Kubernetes v1.5.2 (static) example clusters
-* Upgrade Kubernetes v1.5.3 (self-hosted) example cluster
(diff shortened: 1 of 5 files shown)


aaronlevy:
v0.3.10 doesn't contain v1.5.4. There might not be any problems here, but technically your kubelets shouldn't be a higher version than your control-plane.

I will be cutting another bootkube release shortly which does contain v1.5.4

dghubble:
Planning for the future :sparkles:

dghubble:
The :sparkles: future :sparkles:  has arrived.

aaronlevy:
lgtm. I kinda like the execStartPre approach -- may incorporate that into bootkube.

dghubble:
Props to @ericchiang for that one

Issue #461, opened 2017-03-14
Ignition Partials


nickethier:
I want to start a discussion around template partials. I've been making use of the Ignition append/replace configurations, but it starts to get onerous the more metadata you want to include. One reason for this is that I've got lots of duplicate configuration between my k8s-worker and k8s-controller configurations and I'd like to clean that up so its not so much of a maintenance nightmare.

It is desirable that I can pull out a lot of common duplication into a single template then use `{{ template "foo" . }}` in my main template to pull it in. One idea I had was to add an optional variable to profiles that would specify files you want to pull into the template context before rendering. I've outlined an example below.

I certainly don't thing this replaces the ignition apped/replace functionality. I think thats the first thing users should start with. However as configuration gets more complex it just isn't flexible enough for all use cases. 

I'm willing to put a PR together to make this work.

ref: #322 #230 

profiles/k8s-worker.json:
```
{
  "id": "k8s-worker",
  "name": "Kubernetes Worker",
  "boot": {
    ...
  },
  "ignition_id": "k8s-worker.yaml",
  "partials": ["k8s-common"]
}
```

partials/k8s-common.tmpl (or .yaml):
```
{{ define "k8s-common-units" }}
    - name: k8s-certs@.service
      contents: |
        [Unit]
        Description=Fetch Kubernetes certificate assets
        Requires=network-online.target
        After=network-online.target
        [Service]
        ExecStartPre=/usr/bin/mkdir -p /etc/kubernetes/ssl
        ExecStart=/usr/bin/bash -c "[ -f /etc/kubernetes/ssl/%i ] || curl {{.k8s_cert_endpoint}}/tls/%i -o /etc/kubernetes/ssl/%i"
// Many other units that are shared...
{{ end }}
{{ define "k8s-common-files" }}
  - path: /etc/kubernetes/manifests/kube-proxy.yaml
  ...
{{ end }}
```

ignition/k8s-worker.yaml:
```
---
systemd:
  units:
{{ template "k8s-common-units" . }}
  <worker specific units>
storage:
  files:
{{ template "k8s-common-files" . }}
  <worker specific files>
```

mcuadros:
I found this issue very important, we have very long files with duplicated information such as: network configuration, partition tables, sshd config, etc. We couldn't find any proper way to reuse ignition configuration across different profiles or groups, besides the appends/replace.

The appends/replaces IMHO have the following problems:
- Hard to define, you should create the profile and the groups per ignition append
[...]

PR #462, opened 2017-03-15
ignition, support for include in fuze rederization

mcuadros:
This PR introduce the ability to include partial include files, inside other files, allowing to remove duplicated code:

The concept is similar to the [include function in helm](https://github.com/kubernetes/helm/blob/master/docs/chart_template_guide/named_templates.md#the-include-function), as in helm, also implements an indent function to avoid the indentation problems. 

main.yaml:
```yaml
systemd:
  units:
    - name: {{.service_name}}.service
      enable: true
    - name: {{.uuid}}.service
      enable: true

{{ include "partial.yaml" . | indent 4 }}
```

partial.yaml: 
```yaml
- name: {{.request.query.foo}}.service
  enable: true
  contents: {{.request.raw_query}}
```

If this is something that could be interesting to merge I will be happy to update the documentation.

Related to: https://github.com/coreos/matchbox/issues/461

### DIFF
--- a/matchbox/http/ignition.go
+++ b/matchbox/http/ignition.go
@@ -2,10 +2,10 @@ package http
 
 import (
 	"bytes"
+	"context"
 	"net/http"
 	"strings"
 
-	"context"
 	"github.com/Sirupsen/logrus"
 	fuze "github.com/coreos/fuze/config"
 	ignition "github.com/coreos/ignition/config"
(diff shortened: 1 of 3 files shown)


crawford:
What is this specifically trying to solve that YAML doesn't already have a solution for?

mcuadros:
How can you include a fuze file in fuze file? The problem that I am trying to solve Is at explained at #461

PR #475, opened 2017-04-13
contrib/k8s: Run matchbox on Kubernetes behind Ingress

dghubble:
Demonstrate deploying a public matchbox behind Ingress and verify Terraform apply can update machine configs.

* Show matchbox deployment, service, and TLS secret creation
* Provide an Ingress resource for exposing HTTP and gRPC APIs
* Add note mentioning matchbox can be run publicly if best practices are followed

Note: 
* Guides for the different ways to correctly run Nginx Ingress Controllers on a bare-metal cluster is left for a future document.
* Tutorials on how to use the alpha Terraform plugin are left for a future document

### DIFF
--- a/Documentation/deployment.md
+++ b/Documentation/deployment.md
@@ -1,4 +1,3 @@
-
 # Installation
 
 This guide walks through deploying the `matchbox` service on a Linux host (via RPM, rkt, docker, or binary) or on a Kubernetes cluster.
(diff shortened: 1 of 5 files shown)


Documentation/deployment.md
```diff
@@ -269,24 +268,41 @@ Create machine profiles, groups, or Ignition configs at runtime with `bootcmd` o
...
+
+Verify `http://matchbox.example.com` responds with the text "matchbox" and verify gRPC clients can connect to `matchbox-rpc.example.com:443`.
+
+```sh
+$ curl http://matchbox.example.com
+$ openssl s_client -connect matchbox-rpc.example.com:443 -CAfile ca.crt -cert client.crt -key client.key
+```
 
-## Documentation
+### Kubernetes Notes
```
joshix: both of the notes beneath this heading seem to be about securing secrets and data when exposing the service publicly. Should this heading be something like "Securing Matchbox assets for public service"?

dghubble: The data note is just about persistence because it can be annoying if matchbox forgets configs bc the pod was rescheduled. But available persistent volumes vary so I've left the examples simple. Perhaps "Caveats".

Documentation/deployment.md
```diff
@@ -269,24 +268,41 @@ Create machine profiles, groups, or Ignition configs at runtime with `bootcmd` o
...
 
-View the [documentation](https://github.com/coreos/matchbox#coreos-on-baremetal) for `matchbox` service docs, tutorials, example clusters and Ignition configs, PXE booting guides, or machine lifecycle guides.
+* Secrets: Matchbox **can** be run as a public facing service. However, you **must** follow best practices and avoid writing secret material into machine configs. Instead, load secret materials from an internal secret store.
```
ElijahCaine: Is there some reference, written by us or otherwise, that you can point to that might help people accomplish this?

dghubble: There is not. Basically, this is a statement that matchbox is not a secret store. Its out of scope.

Someday we may get docs on integrating Vault in a recommended way.

Documentation/deployment.md
joshix: Now I get it. Stet -- no change required

dghubble: I'll call them "Operational notes" for now

PR #498, opened 2017-04-25
scripts: get CoreOS version from current release

dannysauer:
If `$2` is not specified, fetch coreos_version string from `current` release version.  This resolves the situation where the get_coreos script has been updated to reflect current ContainerOS release, but there has not been a new Tectonic release to actually include the updated get_coreos script.  And it makes it easier to just run the script periodically without having to look up the new OS release's version.

### DIFF
--- a/scripts/get-coreos
+++ b/scripts/get-coreos
@@ -6,7 +6,13 @@ set -eou pipefail
 GPG=${GPG:-/usr/bin/gpg}
 
 CHANNEL=${1:-"stable"}
-VERSION=${2:-"1298.7.0"}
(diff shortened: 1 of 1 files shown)


dghubble:
Going to pass on this for now, thanks though.

dannysauer:
No problem. I addressed an issue for a self - contained general-purpose updater, which is what I needed but is not this project's need. :)

dghubble:
Yeah, for yourself, you're welcome to edit `VERSION=${2:-"1298.7.0"}` to `VERSION=${2:-"current"}` which should give you what you want more easily.

PR #504, opened 2017-04-27
examples: Update self-hosted Kubernetes to v1.6.2

dghubble:
Updates both the regular and terraform based self-hosted Kubernetes clusters. Jenkins tests only PXE a 3-node bare-metal cluster using the non-terraform flow for now. Terraform testing is manual and I'll report back.


dghubble:
Tested the terraform-based cluster too.

PR #517, opened 2017-05-09
examples/terraform: Add experimental self-hosted etcd option

dghubble:
* Add an option to try experimental self-hosted etcd which uses the etcd-operator to deploy an etcd cluster as pods atop Kubernetes and disables the on-host etcd cluster on controllers and gateway on workers
* When enabled, configure locksmithd to coordinate reboots through self-hosted etcd since there is no on-host etcd to use.

Warning: This feature is still experimental and potentially [dangerous](https://github.com/coreos/tectonic-installer/pull/601#issuecomment-300242512). It is opt-in.

### DIFF
--- a/examples/terraform/modules/profiles/cl/bootkube-controller.yaml.tmpl
+++ b/examples/terraform/modules/profiles/cl/bootkube-controller.yaml.tmpl
@@ -1,6 +1,7 @@
 ---
 systemd:
   units:
+    {{ if eq .etcd_on_host "true" }}
     - name: etcd-member.service
       enable: true
       dropins:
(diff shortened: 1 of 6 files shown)


examples/terraform/modules/profiles/cl/bootkube-controller.yaml.tmpl
```diff
@@ -1,6 +1,7 @@
 ---
 systemd:
   units:
+    {{ if eq .etcd_on_host "true" }}
```
dghubble: Equality comparison was required because terraform always passes strings (rather than booleans). Go template if conditions evaluate string's that are non-zero length to be truthy.

dghubble:
Closes #515

PR #519, opened 2017-05-09
contrib/rpm: Fixing the source URL format

brianredbeard:
Fixing the source URL format to confirm to more normative rpmbuild
standards and to allow for proper use of spectool/rpmspectool.  This
change now produces a proper archive with the name and version number
used.

### DIFF
--- a/contrib/rpm/matchbox.spec
+++ b/contrib/rpm/matchbox.spec
@@ -8,7 +8,8 @@ Release:	2%{?dist}
 Summary:	Network boot and provision CoreOS machines
 License:	ASL 2.0
 URL:		https://%{import_path}
-Source0:	https://%{import_path}/archive/v%{version}.tar.gz
+Source0:        https://%{import_path}/archive/v%{version}/%{name}-%{version}.tar.gz
+
 
 BuildRequires: golang
 BuildRequires: systemd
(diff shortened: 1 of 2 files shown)


contrib/rpm/matchbox.spec
```diff
@@ -8,7 +8,8 @@ Release:	2%{?dist}
 Summary:	Network boot and provision CoreOS machines
 License:	ASL 2.0
 URL:		https://%{import_path}
-Source0:	https://%{import_path}/archive/v%{version}.tar.gz
+Source0:        https://%{import_path}/archive/v%{version}/%{name}-%{version}.tar.gz
```
dghubble: Huh, I didn't realize Github provided both:

* https://github.com/coreos/matchbox/archive/v0.6.0.tar.gz
* https://github.com/coreos/matchbox/archive/v0.6.0/matchbox-0.6.0.tar.gz

dghubble:
So this should work with `spectool -g -R contrib/rpm/matchbox.spec` rather than manual download, that's the idea here right?

dghubble:
Commit message convention is `contrib/rpm: Blah`

brianredbeard:
This does indeed work with `spectool` (or the rewrite of it - `rpmspectool` ).

dghubble:
Thanks!

PR #522, opened 2017-05-12
examples/terraform: Automate terraform-based bootkube-install

### LINKED ISSUE #507: Terraform bootkube-install should use terraform to render and copy assets
I've written a terraform module generates assets exactly like `bootkube render` does. This will allow the terraform-based bootkube-install example to bring up bare-metal infrastructure, generate assets, and secure copy those assets within a single `terraform apply`. Users picking terraform might as well go all the way and use it for all infrastructure state.

https://github.com/dghubble/bootkube-terraform

dghubble:
* Improve (terraform-based) "bootkube-install" bare-metal vanilla self-hosted Kubernetes cluster example
* Use the pinned [bootkube-terraform](https://github.com/dghubble/bootkube-terraform) module to generate the exact same assets that `bootkube render` (binary) would
* Use Terraform to automate the asset generation, kubeconfig copy, and bootkube start (if part of the flow is already using Terraform, we might as well go all the way)

Benefits:
* Users do not need to download the bootkube binary, render assets, or manually copy assets to nodes
* Matchbox continues to never store secrets
* Easier to add alongside existing bare-metal cluster tests if setup is fully automated
* Terraform Module for bare-metal bootkube is available to anyone 

```
source = git::https://github.com/coreos/matchbox.git//examples/terraform/modules/bootkube
```
The manual bootkube-install example cluster for bare-metal still exists under examples.

Closes #507


dghubble:
Isn't working with `experimental_self_hosted_etcd = "true"`. I believe bootstrap apiserver needs to have --etcd-servers=http://10.3.0.15:2379,http://127.0.0.1:12379 and lacks the localhost one currently. Needs a patch before merge.

dghubble:
Fixed, works with and without (default) self-hosted etcd.

dghubble:
Tested with and without self-hosted etcd again. Merging.

PR #527, opened 2017-05-14
Add support for OEM images

enilfodne:
Currently, when using `terraform`, the setup is using a generic image which can cause issues on some platforms. Common problem is the lack of `VM Guest Tools`, when installing on a VMware platform.

This PR adds the ability to fetch OEM images with `get-coreos` and allows a user to specify an oem image in `terraform.tfvars`.

### DIFF
--- a/examples/terraform/bootkube-install/terraform.tfvars.example
+++ b/examples/terraform/bootkube-install/terraform.tfvars.example
@@ -19,4 +19,5 @@ k8s_domain_name = "cluster.example.com"
 asset_dir = "assets"
 
 # Optional
+# container_linux_oem = ""
 # experimental_self_hosted_etcd = "true"
--- a/examples/terraform/etcd3-install/terraform.tfvars.example
+++ b/examples/terraform/etcd3-install/terraform.tfvars.example
@@ -1,3 +1,6 @@
 matchbox_http_endpoint = "http://matchbox.example.com:8080"
 matchbox_rpc_endpoint = "matchbox.example.com:8081"
 # ssh_authorized_key = "ADD ME"
+
+# Optional
+# container_linux_oem = ""
--- a/scripts/get-coreos
+++ b/scripts/get-coreos
@@ -1,6 +1,9 @@
 #!/bin/bash
 # USAGE: ./scripts/get-coreos
 # USAGE: ./scripts/get-coreos channel version dest
+#
+# ENV VARS:
+# - OEM_ID - specify OEM image id to download, alongside the default one
 set -eou pipefail
 
 GPG=${GPG:-/usr/bin/gpg}
@@ -8,6 +11,7 @@ GPG=${GPG:-/usr/bin/gpg}
 CHANNEL=${1:-"stable"}
 VERSION=${2:-"1298.7.0"}
 DEST_DIR=${3:-"$PWD/examples/assets"}
+OEM_ID=${OEM_ID:-""}
 DEST=$DEST_DIR/coreos/$VERSION
 BASE_URL=https://$CHANNEL.release.core-os.net/amd64-usr/$VERSION
 
@@ -20,6 +24,16 @@ fi
 if [ ! -d "$DEST" ]; then
   echo "Creating directory $DEST"
   mkdir -p $DEST
+fi
+
+if [[ -n "${OEM_ID}" ]]; then
+  IMAGE_NAME="coreos_production_${OEM_ID}_image.bin.bz2"
+
+  # check if the oem version exists based on the header response
+  if ! curl -s -I $BASE_URL/$IMAGE_NAME | grep -q -E '^HTTP/[0-9.]+ [23][0-9][0-9]' ; then
+    echo "OEM version not found"
+    exit 1
+  fi
 fi
 
 echo "Downloading CoreOS $CHANNEL $VERSION images and sigs to $DEST"
@@ -46,7 +60,20 @@ curl -# $BASE_URL/coreos_production_image.bin.bz2 -o $DEST/coreos_production_ima
 echo "coreos_production_image.bin.bz2.sig"
 curl -# $BASE_URL/coreos_production_image.bin.bz2.sig -o $DEST/coreos_production_image.bin.bz2.sig
 
+# Install oem image
+if [[ -n "${IMAGE_NAME-}" ]]; then
+  echo $IMAGE_NAME
+  curl -# $BASE_URL/$IMAGE_NAME -o $DEST/$IMAGE_NAME
+  echo "$IMAGE_NAME.sig"
+  curl -# $BASE_URL/$IMAGE_NAME.sig -o $DEST/$IMAGE_NAME.sig
+fi
+
 # verify signatures
 $GPG --verify $DEST/coreos_production_pxe.vmlinuz.sig
 $GPG --verify $DEST/coreos_production_pxe_image.cpio.gz.sig
 $GPG --verify $DEST/coreos_production_image.bin.bz2.sig
+
+# verify oem signature
+if [[ -n "${IMAGE_NAME-}" ]]; then
+  $GPG --verify $DEST/$IMAGE_NAME.sig
+fi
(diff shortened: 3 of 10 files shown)


examples/terraform/bootkube-install/README.md
```diff
@@ -35,8 +35,20 @@ matchbox_rpc_endpoint = "matchbox.example.com:8081"
...
 
+##### Self-hosted etcd
+
 You may set `experimental_self_hosted_etcd = "true"` to deploy "self-hosted" etcd atop Kubernetes instead of running etcd on hosts directly. Warning, this is experimental and potentially dangerous.
 
+##### OEM Customisation
+
+You may set `container_linux_oem = "<oem_id>"` to use a specific oem version as installation base. For example setting `container_linux_oem = "vmware_raw"` will use an installation image (as base) with included Guest Tools.
+
+Popular OEM images: `ami`, `cloudstack`, `digitalocean`, `packet`, `rackspace`, `vmware_raw` and `xen`.
```
dghubble: Bare-Metal is the primary use case and the default "" option is suitable. I see `vmware_raw`, `packet`, and `xen` as plausible options since users can treat those platforms like bare-metal and PXE boot machines there.

The use case for cloud images like `ami` and `digitalocean` seems more nebulous. Can you elaborate on your setup?

Also, let's move this down to the bottom. I don't want to confuse users, most of whom shouldn't need to set this.

examples/terraform/bootkube-install/README.md
```diff
@@ -35,8 +35,20 @@ matchbox_rpc_endpoint = "matchbox.example.com:8081"
...
+
+_Note_: In order to use a specific image, you'll have to fetch it first with ```get-coreos```
```
dghubble: Maybe move this by the `cached-container-linux-install` note. Matchbox itself has no requirement you use its cached assets, its just a convenience in that example profile

examples/terraform/bootkube-install/terraform.tfvars.example
```diff
@@ -1,4 +1,5 @@
 matchbox_http_endpoint = "http://matchbox.example.com:8080"
 matchbox_rpc_endpoint = "matchbox.example.com:8081"
 # ssh_authorized_key = "ADD ME"
+# container_linux_oem = ""
```
dghubble: Place this under # Optional or remove please

examples/terraform/etcd3-install/README.md
```diff
@@ -35,6 +35,16 @@ matchbox_rpc_endpoint = "matchbox.example.com:8081"
 ssh_authorized_key = "ADD ME"
 ```
 
+##### OEM Customisation
```
dghubble: Maybe we carve out a place to have docs for individual profiles. As more example clusters are added, we probably don't want to duplicate. Though I appreciate that you've written docs.

examples/terraform/etcd3-install/terraform.tfvars.example
```diff
@@ -1,3 +1,4 @@
 matchbox_http_endpoint = "http://matchbox.example.com:8080"
 matchbox_rpc_endpoint = "matchbox.example.com:8081"
 # ssh_authorized_key = "ADD ME"
+# container_linux_oem = ""
```
dghubble: Place this under # Optional or remove

scripts/get-coreos
```diff
@@ -1,13 +1,14 @@
...
-# USAGE: ./scripts/get-coreos channel version dest
+# USAGE: ./scripts/get-coreos channel version dest oem_id
 set -eou pipefail
 
 GPG=${GPG:-/usr/bin/gpg}
 
 CHANNEL=${1:-"stable"}
 VERSION=${2:-"1298.7.0"}
 DEST_DIR=${3:-"$PWD/examples/assets"}
+OEM_ID=${4:-""}
```
dghubble: Rather than adding more positional arguments, I think this one needs to be an optional argument

scripts/get-coreos
```diff
@@ -41,12 +54,12 @@ echo "coreos_production_pxe_image.cpio.gz.sig"
 curl -# $BASE_URL/coreos_production_pxe_image.cpio.gz.sig -o $DEST/coreos_production_pxe_image.cpio.gz.sig
 
 # Install image
-echo "coreos_production_image.bin.bz2"
-curl -# $BASE_URL/coreos_production_image.bin.bz2 -o $DEST/coreos_production_image.bin.bz2
-echo "coreos_production_image.bin.bz2.sig"
-curl -# $BASE_URL/coreos_production_image.bin.bz2.sig -o $DEST/coreos_production_image.bin.bz2.sig
+echo $IMAGE_NAME
```
dghubble: Let's always download the default image (even if an oem is specified). I can already see a Tectonic user using an oem option and then being confused when the Installer app doesn't work since it only supports the default image.

examples/terraform/simple-install/groups.tf
```diff
@@ -4,6 +4,7 @@ resource "matchbox_group" "default" {
   profile = "${matchbox_profile.coreos-install.name}"
   # no selector means all machines can be matched
   metadata {
+    container_linux_oem = "${var.container_linux_oem}"
```
dghubble: simple-install is for the getting started tutorial targeted at first time users with bare-metal labs. I'd kinda prefer it be as minimal as possible and the module just have advanced options for those who need them

scripts/get-coreos
enilfodne: Not sure what you mean. Its an optional argument, omitting `oem_id` will not affect the functionality in any way.

enilfodne:
Rebased and followed-through on your feedback (thanks), i've removed the `README` updates as it seems they were out-of-place and confusing.

examples/terraform/bootkube-install/README.md
enilfodne: I'm using VMware's platform, but i listed all possible images (at the time of writing) as they maybe of use to someone experimenting.

scripts/get-coreos
dghubble: Positional as in this is locked in as the 4th argument. Optional argument would be `-oem_id=""`.

If a 5th argument was added for another optional parameter, you'd have to pass something for the 4th. It starts to get unwieldy.

dghubble: In the short term, just set OEM_ID="${OEM_ID:-""}"} so people can set the environment variable before running the script.

enilfodne: Sorry, drew a blank on this one. Fixed

examples/terraform/modules/bootkube/bootkube.tf
```diff
@@ -11,5 +11,6 @@ module "bootkube" {
   kube_apiserver_service_ip     = "${var.k8s_apiserver_service_ip}"
   kube_dns_service_ip           = "${var.k8s_dns_service_ip}"
   kube_etcd_service_ip          = "${var.k8s_etcd_service_ip}"
+  container_linux_oem           = "${var.container_linux_oem}"
```
dghubble: The bootkube-terraform module just renders asset files, it doesn't take this value.

scripts/README.md
```diff
@@ -6,7 +6,7 @@
 Run the `get-coreos` script to download CoreOS images, verify them, and move them into `examples/assets`.
 
     ./scripts/get-coreos
-    ./scripts/get-coreos channel version
+    ./scripts/get-coreos channel version oem_id
```
dghubble: remove the oem_id position argument from here

scripts/get-coreos
dghubble: ^

dghubble:
LGTM, can you squash this to a single commit which has the commit format in the style guide? e.g. "examples/terraform: Add support for OEM images"

dghubble:
I'm not sure why Travis won't test this PR, but no matter.

PR modifies cluster examples and looks good to test. Pushed to an origin branch which will boot the examples clusters in parallel. https://github.com/coreos/matchbox/tree/support-oems

dghubble:
Passed everything, merged. Thanks!

PR #528, opened 2017-05-15
examples: Add NoSchedule taint to bootkube controllers

dghubble:
I noticed a difference btw these examples and bootkube examples is that bootkube disables workload scheduling on controllers as of Kubernetes v1.6. Let's do the same.

*Note: This means clusters must have at least 1 controller and 1 worker to be useful.*


dghubble:
That should be fine, matchbox cluster tests PXE 1 controller and 2 workers. :+1:

Issue #523, opened 2017-05-12
Fix bootkube cluster testing flakiness


dghubble:
The smoke tests for the bootkube cluster wait until nodes and pods are ready / running (loop and tolerate failures and blips) and then attempt to get the final list of nodes and pods to the logs (failures not allowed). I've noticed through a number of test flakes that the control plane is still stabilizing in those first few moments so sometimes the get nodes or get pods fails.

https://github.com/coreos/matchbox/blob/master/tests/smoke/bootkube#L50

Waiting until all nodes and pods are running is not a sufficient way to determine when the cluster is truly "ready" for us to perform checks which are required to succeed. Come up with a better way to assess when the cluster is "ready".

dghubble:
Actually, these tests only wait for nodes to be ready. I think it was the Tectonic tests I made to wait on both nodes and pods so that might be an improvement.

https://github.com/coreos/tectonic-installer/blob/master/installer/tests/scripts/bare-metal/up-down.sh#L77

dghubble:
To clarify, this flakiness is in the tests. bootkube itself seems to be bringing up the cluster successfully and flakes there are not expected.

dghubble:
Fixed by https://github.com/coreos/matchbox/pull/542/commits/05925036528858e6b20284dc1c5414e16aa561c4

PR #535, opened 2017-05-19
tests: Add cluster tests for bootkube-install (terraform-based)

dghubble:
Matchbox origin branches now test PXE booting and provisioning the following clusters on each commit:
* 3-node etcd
* 3-node etcd (terraform based)
* 3-node static Kubernetes
* 3-node self-hosted Kubernetes
* 3-node self-hosted Kubernetes (terraform-based, upstream for Tectonic)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,10 +46,22 @@ parallel (
   "etcd3-terraform": {
     node('fedora && bare-metal') {
       stage('etcd3-terraform') {
-        timeout(time:8, unit:'MINUTES') {
+        timeout(time:10, unit:'MINUTES') {
           checkout scm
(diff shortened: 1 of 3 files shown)


dghubble:
These clusters bootstrap successfully most of the time, and other times hit flakes in underlying components or in the testing strategy itself.

Overall these function as smoke tests across 5 supported cluster types. A green test provides confidence in a commit while a failure usually highlights unreliability in some cluster bring-up, since these tests are subject to the flakes in every subcomponent. Repeated failures on the same cluster in the grid indicates a true regression. Reducing overall flakes is tracked in #523

dghubble:
screenshot from 2017-05-19 11-28-38

PR #542, opened 2017-05-22
tests: Temporarily disable bootkube (terraform-based) cluster testing

### LINKED ISSUE #541: Tests are quite flaky and require several retries
Testing 5 difference clusters means any flake in any underlying cluster, component, or test environment fails the build. Investigate these flakes try to solve them or throw out clusters that are lower priority - the current flake rate is too high.

Individually running each cluster test locally succeeds.

dghubble:
The bootkube-terraform test leaves behind terraform PIDs which run apply and interfere with terraform-based tests. This leaves the fleet of executors in a bad state. Temporarily disable these tests and roll the executors.

Closes #541 #523

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,16 +55,4 @@ parallel (
       }
     }
   },
-  "bootkube-terraform": {
-    node('fedora && bare-metal') {
-      stage('bootkube-terraform') {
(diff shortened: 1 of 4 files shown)


dghubble:
These are passing nicely, much more reliably.

PR #544, opened 2017-05-22
Remove static Kubernetes and rktnetes example clusters

### LINKED ISSUE #539: Remove static Kubernetes / rktnetes example clusters
These example clusters still deploy Kubernetes 1.5 and aren't being actively developed. Instead, the self-hosted style is recommended. Longer discussion in #512 

cc @pbx0 @robszumski

dghubble:
* Static Kubernetes / rktnetes examples are no longer going to be maintained by this repo or upgraded to Kubernetes v1.6. This is not considered a deprecation bc the reference clusters are examples.
* Remove static Kubernetes cluster examples so users don't accidentally choose it
* Self-hosted Kubernetes (bootkube) is now the standard recommended Kubernetes cluster configuration

cc @pbx0 @robszumski 
Closes #539


robszumski:
Sounds good to me

PR #540, opened 2017-05-20
Update CoreOS to the latest available version

enilfodne:
Based on [CoreOS Release Page](https://coreos.com/releases/), the latest stable version is `1353.7.0`.

#### Reason to update

A newly created cluster randomly updated a few machines to `1353.7.0` and left some on the current version.

#### Updated components

| component | cur. version | upd. version |
| --- | --- | --- |
| kernel | 4.9.16 | 4.9.24 |
| rkt | 1.21.0 | [1.25.0](https://github.com/rkt/rkt/blob/master/CHANGELOG.md#1250) |
| systemd | 231 | [233](https://github.com/systemd/systemd/blob/master/NEWS) |

#### Known Issues

Testing so far has revealed no incompatibility or issues, compared to `1298.7.0`.

PR #546, opened 2017-05-23
Bump Container Linux version to stable 1353.7.0

dghubble:
Rebases #540 and runs cluster tests. Minor fixes.

### DIFF
--- a/Documentation/api.md
+++ b/Documentation/api.md
@@ -39,8 +39,8 @@ GET http://matchbox.foo/ipxe?label=value
 
 ```
 #!ipxe
-kernel /assets/coreos/1298.7.0/coreos_production_pxe.vmlinuz coreos.config.url=http://matchbox.foo:8080/ignition?uuid=${uuid}&mac=${mac:hexhyp} coreos.first_boot=1 coreos.autologin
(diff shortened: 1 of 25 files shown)


dghubble:
It might be an option in the future if done right. But there are concerns.

Matchbox objects or Terraform configs which use "current" no longer completely declare how a cluster should be created. It depends on the "current" image in the assets cache in the environment and users must override that image often or keep track of when different machines were deployed. Its the same reason Kubernetes manifests generally shouldn't specify a "latest" image. Regular users probably need a symlink setup where discrete versions are downloaded, but current points to the highest version.

The second matter is CI cluster testing. We could setup CI to always download current in every test so we're guaranteed each test runs at the latest release as of the test timestamp. There is a risk that non-deterministic builds will result or a CL release will break tests, but with stable that may be a lower risk than it was in the past.

dghubble:
Also, we should carve out separate folders for alpha, beta, and stable in the assets cache. Right now, the current of each channel will override each other which doesn't work well for multi-channel environments without modifying examples. This was avoided by using the version numbers.

For example `/assets/coreos/VERSION` -> `/assets/coreos/CHANNEL/VERSION`

PR #548, opened 2017-05-23
examples/terraform: Add tfvars showing multi-controller case

### LINKED ISSUE #398: Add a multi-controller Kubernetes example cluster
Straight forward extension of the single-controller works fine, it was implemented in Tectonic some time ago. It might be useful to show open-source people an example?

dghubble:
Closes #398


dghubble:
Btw, I tested this with and without self-hosted etcd starting from multiple controllers. @aaronlevy I believe we were talking about this, but I could be thinking of a different conversation.

Issue #550, opened 2017-05-24
CentOS 7 RPM availability


dghubble:
Moving question from #525

> Agreed. I just tried installing matchbox onto CentOS and the instructions were either unclear or didn't work.
Example: https://coreos.com/matchbox/docs/latest/deployment.html#rpm-based-distro
RPM-based distro
On an RPM-based provisioner, install the matchbox RPM from the Copr repository using dnf or yum.
dnf copr enable @CoreOS/matchbox
dnf install matchbox
CentOS is an RPM-based distro but it doesn't use dnf. What are the yum commands needed?
Thanks!

> FYI - I was able to get it working, but it wasn't obvious from the instructions that the systemd copied service was "matchbox-local" and not just "matchbox", therefore a systemctl "start matchbox-local" is what is needed to be executed to start the service

cc @nsaustin

dghubble:
Good point. In the past, Matchbox's Copr repo used to provide a CentOS 7 RPM which could be installed via `yum`. Matchbox (as of v0.5.0) requires Go 1.7 or higher and CentOS comes with Go 1.6. The CentOS RPM was dropped.

Short term, our docs should only mention `dnf` as only Fedora 24 and 25 are new enough. Instead, downloading a release binary and following the instructions for "general linux" is the best option for CentOS.

Longer term, I'd like to improve this story for those using CentOS.

brianredbeard:
This can be built on top of CentOS/RHEL using Software Collections.  We could potentially make that repo a `BuildRequires` option in the spec.

PR #551, opened 2017-05-24
CHANGES.md: Prepare for a v0.6.1 docs point release

dghubble:
* Release will allow docs team to update docs (also closes https://github.com/coreos/tectonic-installer/issues/672)
* Release branch will be used to avoid deprecation changes being included in the point release
* Bumping Go to 1.8.3 for release images and binaries (#552)
* I'm not planning an RPM release. Docs and Go build version are independent from RPM releases so there are effectively no changes to the package.
* Clarifies #550 

Branch to be tagged:
https://github.com/coreos/matchbox/tree/release-v0.6.1

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,12 @@ Notable changes between releases.
 
 * Remove pixiecore support (deprecated in v0.5.0)
 
+## v0.6.1 (2017-05-25)
+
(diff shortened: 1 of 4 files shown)


dghubble:
Maybe we bump to Go 1.8.3 as part of this point release as well.

Issue #486, opened 2017-04-18
Jenkins smoke tests should build/test matchbox container


dghubble:
Smoke tests PXE boot etcd3, k8s, and bootkube clusters, but this currently only tests the cluster examples. We should build matchbox on every origin PR, create a container, and PXE boot clusters against that matchbox container to cover changes to the matchbox service.

dghubble:
The testing infrastructure is no longer available for this. Plus, the core functionality of Matchbox the application is tested through unit tests. Spinning up a Kubernetes cluster is a project in its own right, so it shouldn't be in the test setup for Matchbox.

PR #555, opened 2017-05-26
Jenkinsfile: Switch to declarative-style Jenkins pipeline

dghubble:
Use the newer, declarative pipeline style.

Related to #486, but does not close.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,6 +17,7 @@ pipeline {
                 sh '''#!/bin/bash -e
                 export ASSETS_DIR=~/assets; ./tests/smoke/etcd3
                 '''
+                deleteDir()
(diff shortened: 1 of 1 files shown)


dghubble:
Kinda a bummer that with declarative pipelines, `parallel` must be within a stage and stages can't be nested. That means the Jenkins UI shows one box for all of "Cluster Tests" where it used to breakout the logs, time taken, and color indicator for each scripted stage (e.g. etcd3, bootkube, etcd3-terraform).

https://issues.jenkins-ci.org/browse/JENKINS-41334 seems to maybe track this because being able to declare that top level states should execute in parallel would solve this problem.

dghubble:
Thanks

PR #564, opened 2017-06-05
matchbox: Remove Profile cmdline map field

### LINKED ISSUE #533: Remove Profile cmdline argument
The Matchbox Profile type has both `cmdline` (a `map[string]string`) and `args` (a `[]string`) to represent kernel command line arguments. Cmdline was a flaw in the initial implementation which has been marked deprecated since December 2016. https://github.com/coreos/matchbox/pull/396

Remove the `cmdline` field. This will not affect the Tectonic (terraform-based) installer. The older Tectonic installer (non-terraform based) will not be compatible with matchbox v0.7+ unless its updated with the same fix.

cc @sym3tri @Quentin-M

dghubble:
Remove Profile kernel args `Cmdline` map (deprecated in v0.5.0), use kernel `Args` slice instead

Details in #533. Closes #533


dghubble:
Executor connectivity was fixed.

PR #571, opened 2017-06-12
examples/terraform/modules: Add outputs.tf with kubeconfig

dghubble:
Enables module users to use the cluster's kubeconfig in their own, add-on infrastructure declarations, such as adding diskless workers.


dghubble:
I rely on this and know its working.

PR #582, opened 2017-06-16
Adding init script for matchbox

nmaupu:
Added a rapid implementation of an init script useful for some systems that does not provide systemd yet.


dghubble:
I don't think this can stand alone, its needs:

* Add a section to the deployment.md guide (also mention customization and logging)
* Define at least one pre-systemd era setup this has been used on
* Init script assumes `--address 0.0.0.0:8080` should be used

There is the additional question of how this will be maintained over time, it seems likely to go unsupported.

PR #591, opened 2017-06-26
examples: Use etcd client certs in locksmithd dropin

### LINKED ISSUE #590: CL auto-upgrades will not occur with self-hosted etcd and TLS
Following the [Kubneretes v1.6.6 upgrade](https://github.com/coreos/matchbox/pull/585) upgrade, when experimental self-hosted etcd is enabled, auto-updates are broken. Self-hosted etcd now always uses TLS in our examples, just like on-host etcd does. The `locksmithd.service` needs to be configured to use the etcd client credentials in both cases. I missed this in the upgrade.

```
# bad
journalctl -f -u locksmithd
Unlocking old locks failed: error setting up lock: Error initializing etcd client: client: etcd cluster is unavailable or misconfigured. Retrying in 5m0s.

# happy
Jun 26 21:31:18 node3.example.com locksmithd[818]: LastCheckedTime=0 Progress=0 CurrentOperation="UPDATE_STATUS_CHECKING_FOR_UPDATE" NewVersion=0.0.0 NewSize=0
Jun 26 21:31:19 node3.example.com locksmithd[818]: LastCheckedTime=1498512679 Progress=0 CurrentOperation="UPDATE_STATUS_IDLE" NewVersion=0.0.0 NewSize=0
```

Clusters created at 6f02107448cac30bfb0a4200933c1712f1673d9e and ff06990edb96de2c80057030fbfb098edaee2bbc (merge commit) are affected and should be re-provisioned or you may apply the patch manually. In other words, if your terraform config contains:

```
module "mycluster" {
  source = "git::https://github.com/coreos/matchbox.git//examples/terraform/modules/bootkube?ref=6f02107448cac30bfb0a4200933c1712f1673d9e"
```

you need to fix the locksmithd configuration.

```
sudo systemctl edit locksmithd.service
<- see PR for edit
sudo systemctl daemon-reload
sudo systemctl restart locksmithd
```

This change did not make it into Tectonic, because Tectonic already uses CLUO. We'd like to start using CLUO here as well. #587

dghubble:
* Fixes a regression introduced in 6f02107 which upgraded to Kubernetes v1.6.6 and added self-hosted etcd with TLS
* Both on-host and self-hosted etcd now require clients to use TLS client certs so locksmithd

Closes #590

### DIFF
--- a/examples/terraform/modules/profiles/cl/bootkube-controller.yaml.tmpl
+++ b/examples/terraform/modules/profiles/cl/bootkube-controller.yaml.tmpl
@@ -32,12 +32,12 @@ systemd:
           contents: |
             [Service]
             Environment="REBOOT_STRATEGY=etcd-lock"
-            {{ if eq .etcd_on_host "false" -}}
-            Environment="LOCKSMITHD_ENDPOINT=http://{{.k8s_etcd_service_ip}}:2379"
-            {{ else }}
             Environment="LOCKSMITHD_ETCD_CAFILE=/etc/ssl/etcd/etcd-ca.crt"
             Environment="LOCKSMITHD_ETCD_CERTFILE=/etc/ssl/etcd/etcd-client.crt"
             Environment="LOCKSMITHD_ETCD_KEYFILE=/etc/ssl/etcd/etcd-client.key"
+            {{ if eq .etcd_on_host "false" -}}
+            Environment="LOCKSMITHD_ENDPOINT=http://{{.k8s_etcd_service_ip}}:2379"
+            {{ else }}
             Environment="LOCKSMITHD_ENDPOINT={{.etcd_endpoints}}"
             {{ end }}
     - name: kubelet.path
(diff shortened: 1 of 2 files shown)


dghubble:
For clarity (Github UI makes this hard to notice), this change ensures the ETCD_CA, ETCD_CERTFILE, and ETCD_KEYFILE are always a part of the dropin, rather than in the conditional.

```
    - name: locksmithd.service
      dropins:
        - name: 40-etcd-lock.conf
          contents: |
            [Service]
            Environment="REBOOT_STRATEGY=etcd-lock"
            Environment="LOCKSMITHD_ETCD_CAFILE=/etc/ssl/etcd/etcd-ca.crt"
            Environment="LOCKSMITHD_ETCD_CERTFILE=/etc/ssl/etcd/etcd-client.crt"
            Environment="LOCKSMITHD_ETCD_KEYFILE=/etc/ssl/etcd/etcd-client.key"
            {{ if eq .etcd_on_host "false" -}}
            Environment="LOCKSMITHD_ENDPOINT=http://{{.k8s_etcd_service_ip}}:2379"
            {{ else }}
            Environment="LOCKSMITHD_ENDPOINT={{.etcd_endpoints}}"
            {{ end }}
```

PR #596, opened 2017-06-30
examples: include 'curl' retries

euank:
`After=network-online.target` *should* mean this isn't needed in most
cases, but per
https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/, the
definition of "network-online" is a little shaky.

Regardless, being a little more resilient to network flakes and races is
a good thing. The count of `10` was arbitrarily chosen.

No particular testing done.

### DIFF
--- a/examples/ignition/install-reboot.yaml
+++ b/examples/ignition/install-reboot.yaml
@@ -20,7 +20,7 @@ storage:
       contents:
         inline: |
           #!/bin/bash -ex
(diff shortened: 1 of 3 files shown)


dghubble:
If you create your branch on origin, Jenkins will create all the different clusters for you.

dghubble:
Originally, the `installer.service` was written to restart on-failure and the script has -e. This profile is used for usually unattended installs so there isn't much else to do besides retry anyway. We might revisit that, there were some concerns before. cc @crawford

PR #597, opened 2017-07-07
Documentation tweaks

ivy:
I made a small fix to pluralize "templates" where it was previously written:

> [...] to different profiles and template in machine-specific values.

Additionally, I added syntax highlighting for the code example that follows.


dghubble:
Thanks!

PR #598, opened 2017-07-07
Remove invalid port in grub.cfg example

bgeesaman:
docs: Remove invalid port in grub.cfg example

Grub2's http.mod doesn't allow for a port specification (assumes tcp/80), so it needs to be removed and a note about moving matchbox to listen on port 80.  Also, I added the $net_default_mac grub variable to be useful for profile selectors.

### DIFF
--- a/Documentation/network-setup.md
+++ b/Documentation/network-setup.md
@@ -219,10 +219,17 @@ Grub can be used to delegate as well.
 
 /var/lib/tftpboot/boot/grub/grub.cfg:
 ```ini
-insmod i386-pc/http.mod
-set root=http,matchbox.foo:8080
+insmod http
+set root=http,matchbox.foo
 configfile /grub
 ```
+or, if you want to send the system's MAC address to matchbox for use in a profile selector:
+```ini
+insmod http
+set root=http,matchbox.foo
+configfile /grub?mac=$net_default_mac
+```
+*Note: Grub2's http module does not accept a port specification, so ensure that matchbox is running on port 80 instead of 8080 in this case*
 
 Make sure to replace variables in the example config files; instead of iPXE variables, use GRUB variables. Check the [GRUB2 manual](https://www.gnu.org/software/grub/manual/grub.html#Network).
 


mischief:
AFAICT, it is possible to pass ports.

see https://github.com/coreos/grub/blob/master/grub-core/net/http.c#L393 where port is parsed via https://github.com/coreos/grub/blob/master/grub-core/net/net.c#L1258

what makes you think otherwise?

bgeesaman:
Ah, it appears my use of alpine:latest was my downfall.  Here's its version of grub2 [http.c](http://git.savannah.gnu.org/gitweb/?p=grub.git;a=blob;f=grub-core/net/http.c;h=5aa4ad3befc15ba427797c7f1454d8a322d51b2f;hb=HEAD)  Switching to a "normal" base image fixes this issue.  Please disregard. Thanks.

jumanjiman:
@bgeesaman what do you mean by "normal" base image?

bgeesaman:
Sorry for that, it was a late night.  Didn't mean to slight alpine there.  I mean a base image (phusion/baseimage -> Ubuntu 16.04.x, specifically) that has a grub2 package with a similar codebase and features as the coreos/grub repo built-in.  The source of the grub package in alpine doesn't have this non-standard http port feature.

Alpine's http.c line 393](http://git.savannah.gnu.org/gitweb/?p=grub.git;a=blob;f=grub-core/net/http.c;h=5aa4ad3befc15ba427797c7f1454d8a322d51b2f;hb=HEAD)

vs

[CoreOS/Ubuntu/others](https://github.com/coreos/grub/blob/master/grub-core/net/http.c#L393)

jumanjiman:
i meandered down a path to compare grub sources and patches. 😁 

[alpine grub](https://git.alpinelinux.org/cgit/aports/tree/main/grub?h=master) is mostly vanilla and builds `http.c` as-is from upstream grub (iow unpatched and therefore cannot specify non-default port).

The coreos patch for non-default port was [discussed in grub-devel](http://lists.gnu.org/archive/html/grub-devel/2017-01/msg00024.html) but not merged.

Ubuntu includes their own patch (`patches/net_read_bracketed_ipv6_addr.patch`) as part of https://launchpad.net/ubuntu/+source/grub2/2.02~beta2-36ubuntu3 that enables to specify a port, and this version is what you get with image [phusion/baseimage](https://hub.docker.com/r/phusion/baseimage/).

Issue #602, opened 2017-07-11
Update Kubernetes clusters to use bootkube v0.5.0


dghubble:
https://github.com/kubernetes-incubator/bootkube/releases

Issue #601, opened 2017-07-11
Fix bare-metal cluster tests to restore coverage


dghubble:
Some time during the last week (was on vacation), bare-metal cluster tests stopped working. There have been no changes in the cluster definitions, bare-metal executors, or Matchbox source.

Repeated test runs show that that either of the two Kubernetes clusters fails randomly, but given a very long time to run (~40 minutes), both can succeed. Investigating the bare-metal executors in the lab which PXE boot VMs (cluster under test) and getting into one of the launched nodes, it appears cluster bring-up is crawling at the `quay.io/coreos/hyperkube` download step. This would explain why the actual steps of cluster automation seem to be working and why etcd clusters aren't affected (tiny image). It seems there is some change in the network environment at play, perhaps network throttling or QoS limitations.

I'm manually testing the reference clusters now to be sure the definitions themselves are ok.

dghubble:
This masked the introduction of a testing regression.

Updating to bootkube v0.5.0 changed the organization of etcd certificates. This was reflected in documentation (which is why real clusters worked correctly in all manual testing and deployment), but not in the `bootkube` cluster test.

dghubble:
Tests seem to be passing fairly reliably again. We've made no changes in the lab environment or to tests - the duration seems based on a variety of factors and image download speeds depend greatly on the time of day unfortunately. Times vary between 20 minutes and 40 minutes so we'll keep the lenient timeouts for now.

For now, if one of the clusters times out, you may retry. If it occurs repeatedly, especially for the same cluster type, its worth investigating.

PR #604, opened 2017-07-13
examples: Update terraform Kubernetes to use bootkube v0.5.0

dghubble:
* [Render changes](https://github.com/dghubble/bootkube-terraform/pull/5) corresponding to bootkube v0.5.0
* Manually tested cluster bootstrapping, with/without self-hosted etcd, and locksmithd updating the master and recovery. Looks good. Jenkins cluster tests may or may not pass, see #601 

Last part to #602


dghubble:
*Sigh* Manually testing with experimental self-hosted etcd enabled, this doesn't bootstrap. Investigating.

PR #607, opened 2017-07-14
scripts/devnet: set -e

kans:


dghubble:
This seems to instant fail the Jenkins cluster tests for all clusters.

`./script/devnet destroy` is part of `cleanup` which is called at the beginning and end of test runs so as not to assume the prior test or environment cleaned up properly. If the environment is already clean, deletes fail so this script was allowed to run to completion regardless. I think that's the cause.

PR #487, opened 2017-04-19
Update docs, changelog, and scripts for dnsmasq:v0.4.0

dghubble:
* [quay.io/coreos/dnsmasq](https://quay.io/repository/coreos/dnsmasq) v0.4.0 is available
    * dnsmasq package version 2.76
   * Rebuilds with alpine:3.5 base image to receive patches
    * Updates CoreOS `grub.efi` to be recent (stable, 1298.7.0)
* Update docs `coreos.com/dnsmasq` ACI references to `quay.io/coreos/dnsmasq`. Specify v0.4.0 when appropriate
* Update Jenkins CI to use the new dnsmasq:v0.4.0 image

PR #608, opened 2017-07-14
Switch Kubernetes clusters from locksmith to Container Linux Update Operator

dghubble:
* Use Container Linux Update Operator to coordinate reboots in Kubernetes clusters. Mask locksmithd.
* Switch Kubernetes (non-terraform) and Kubernetes (terraform) to use CLUO
* Add a docs page about cluster addons to be used by both flows
* Terraform modules `bootkube` and `profiles` (Kubernetes) now disable locksmithd. Those using the modules directly should take action to deploy CLUO (change).

Closes #487 and  https://github.com/dghubble/bootkube-terraform/pull/4

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,10 @@ Notable changes between releases.
 ### Examples / Modules
 
 * Upgrade Kubernetes v1.6.6 example clusters
-* Kubernetes examples clusters enable etcd TLS (unless experimental self-hosted etcd is enabled)
+* Kubernetes examples clusters enable etcd TLS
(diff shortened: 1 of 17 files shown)


dghubble:
A few caveats for known issues:

* If you use the `bootkube` example (not `bootkube-install`) which is a PXE booted "live" cluster, you'll see CLUO agent crash loop per https://github.com/coreos/container-linux-update-operator/issues/114.

* The message of the day will be misleading and claim the update strategy is No Reboots. See https://github.com/coreos/container-linux-update-operator/issues/88

dghubble:
I've tested the various examples as well. All fine. The Jenkins cluster tests have been having issues since last week, tracked in #601.

PR #609, opened 2017-07-15
check gpg-signatures before downloading, minimizing downloads

strowi:
Before the files would be downloaded every time the scripts was being run.
Now the gpg-signature will be downloaded and checked before downloading. 
Only if the gpg verification fails downloading starts.


dghubble:
This verifies the signature as a way of checking for the file existing and then signatures are checked again at the bottom. Why not check for the pre-existing file?

strowi:
We had some trouble with corrupt downloads, which a simple "test -f" wouldn't detect. 
Only checking the signature at the end would require (manual) deletion and restart of the download.

Since the signature wil be downloaded any way, i thought it would be easier to check the signature upfront too, and if it fails autmatically download the file.

dghubble:
Closing per #641

PR #610, opened 2017-07-17
variables have to be exported

celevra:


dghubble:
This shouldn't be needed

celevra:
In bash its needed
Only if you put it infront of the command, you can use it without export


<http://www.9folders.com/>
________________________________
Von: Dalton Hubble <notifications@github.com>
Gesendet: 17.07.2017 18:43
An: coreos/matchbox
Cc: Philipp Zeitschel; Author
Betreff: Re: [coreos/matchbox] variables have to be exported (#610)


This shouldn't be needed

-

celevra:
Kubectl is a subprocess and can't read my variables

https://stackoverflow.com/questions/1158091/defining-a-variable-with-or-without-export

<http://www.9folders.com/>
________________________________
Von: Dalton Hubble <notifications@github.com>
Gesendet: 17.07.2017 18:43
An: coreos/matchbox
Cc: Philipp Zeitschel; Author
Betreff: Re: [coreos/matchbox] variables have to be exported (#610)


This shouldn't be needed

-

dghubble:
Sure, looks like the local env is setup differently to not require this. Can you update this throughout the docs?

dghubble:
See #613

PR #611, opened 2017-07-17
tests/smoke: Fix etcd certs distribution in bootkube test

dghubble:
* Introduced in ce3154cae962e868faa2b247d856e004eb8545a7
* Masked by larger-scale timeouts / issues in the testing env

Related to #601


dghubble:
:raised_hands:

PR #615, opened 2017-07-19
Documentation: Refresh Kubernetes in-place upgrade doc

### LINKED ISSUE #612: Update the Kubernetes rolling upgrade guide
The Kubernetes manual rolling upgrade docs are out of date and should be updated.

- [x] Bump Kubernetes component numbers from 1.4 to 1.6
- [x] Daemonsets allow rolling upgrades now. Improve docs
- [x] Kubelet is run on-host as a systemd unit. Remove mentions of it being a pod.
- [x] Verify the flow still works on a modern 1.6.X cluster

*Note: This does **not** mean this repo is supporting upgrading from Kubernetes 1.4 to 1.6 (architecture changed greatly). That's not the responsibility of this repo. It means we're showing how to try to rolling upgrade when using a relatively current Kubernetes - if it fails the admin must be ready to nuke and reprovision. If you want guarantees and contracts see Tectonic.*

User encountered some issues going through it recently: https://github.com/coreos/matchbox/issues/588

cc @aaronlevy @kbrwn @crawford @andrewwebber

dghubble:
I've verified these docs doing an in-place update of a bare-metal cluster. Notable changes:

* Daemonsets support rolling upgrades so just edit and wait!
* Kubelet is no longer run as a pod so editing the env file on each host is required
* Added new section calling about how Tectonic users should NOT follow this

cc @kbrwn @andrewwebber
Closes #612

### DIFF
--- a/Documentation/bootkube-upgrades.md
+++ b/Documentation/bootkube-upgrades.md
@@ -1,9 +1,16 @@
-
 # Upgrading self-hosted Kubernetes
 
-[Self-hosted](bootkube.md) Kubernetes clusters schedule Kubernetes components such as the apiserver, kubelet, scheduler, and controller-manager as pods like other applications (except with node selectors). This allows Kubernetes level operations to be performed to upgrade clusters in place, rather than by re-provisioning.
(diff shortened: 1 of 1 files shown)


dghubble:
also cc @robszumski

PR #616, opened 2017-07-20
examples: Install clusters at Container Linux 1409.7.0 (stable)

dghubble:
* PXE boot and/or install using Container Linux at stable 1409.7.0 by default. We bump this every so often just to stay relatively current.


dghubble:
All clusters pass.

PR #618, opened 2017-07-21
examples/addonts: Update CLUO from v0.2.1 to v0.2.2

dghubble:
See https://github.com/coreos/container-linux-update-operator/pull/120


sdemos:
Should this also include the update strategy?

dghubble:
Yeah, lemme add that.

PR #619, opened 2017-07-21
add ignition append to bootkube terraform module

bcg62:
Currently there doesn't seem to be an easy way to inject additional config around the bootkube modules using terraform.

In most enterprise settings there are common configurations which need to get applied to each node regardless what their profile/group is. IE:

-  Network/bonding

- http proxy

- in house CA

- timezone

I'm sure there are many more examples but this would allow adding on without having to manually manage the controller/worker configs outside of the module and it also reduces replication.


dghubble:
I'm not keen to encourage the append pattern for those using Terraform. Either style is ok, but increasingly it is valuable to template configurations early, before writing them to Matchbox or having the machine fetch configs. Matchbox `metadata` and Ignition's `append` features are less relevant in this scenario. Here's [a PR](https://github.com/coreos/matchbox/commit/395494c1d93198f90007a4a441b0bfc0be036085) that moved from matchbox templating to Terraform templating along those lines. In practice, it means:
1) Errors get surfaced when running `terraform apply` rather than when the machine boots and makes network calls
2) Matchbox users benefit from the Terraform ecosystem of templating options (everybody likes their own templating sol)

We do need to work on recommending how you can use our Terraform modules, but customize the configs. Likely improving [outputs.tf](https://github.com/coreos/matchbox/blob/master/examples/terraform/modules/profiles/outputs.tf) for use with [terraform-provider-ct](https://github.com/coreos/terraform-provider-ct), a module which validates Container Linux configs. That would allow appending with any tools you like and still provide the same level of validation.

bcg62:
I can submit separate pull requests to templatize my additional configurations listed above if that is more desirable but i could forsee it bloating quickly especially when trying to account for all the different networkd options someone may wish to use

dghubble:
Filed #622

PR #623, opened 2017-07-26
Add handling of current to get-coreos script

coresolve:
* Adds logic to get-coreos script such that you can provide it with:
```
get-coreos stable current .
```

and it will place the downloaded image into a directory named for the
version. This helps with tectonic as tectonic expects that the
downloaded image is named for the version rather than "current"


dghubble:
Why not just write your own scripts? #641

dghubble:
Closing per #641

Issue #620, opened 2017-07-23
dnsmasq image manifest only lists 53/tcp


vip8439:
The image manifest for quay.io/coreos/dnsmasq only lists port 53/tcp with name port-53. This prevents using `rkt run --port=` to bind other ports from the running dnsmasq instance to a particular host IP. Example: 53/udp, 67/udp, and 69/udp

```{
	"acKind": "ImageManifest",
	"acVersion": "0.6.1",
	"name": "quay.io/coreos/dnsmasq",
	"labels": [
		{
			"name": "version",
			"value": "latest"
		},
		{
			"name": "arch",
			"value": "amd64"
		},
		{
			"name": "os",
			"value": "linux"
		}
	],
	"app": {
		"exec": [
			"/usr/sbin/dnsmasq"
		],
		"user": "root",
		"group": "root",
		"workingDirectory": "/",
		"environment": [
			{
				"name": "PATH",
				"value": "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
			}
		],
		"ports": [
			{
				"name": "port-53",
				"protocol": "tcp",
				"port": 53,
				"count": 1,
				"socketActivated": false
			}
		]
	}
}```

dghubble:
Should be solved in quay.io/coreos/dnsmasq:v0.4.1

vip8439:
Verified new ports are listed in the image manifest.

PR #625, opened 2017-07-31
contrib/dnsmasq: Bump dnsmasq image to v0.4.1

dghubble:
* Update from alpine:3.5 to alpine:3.6
* List ports 67 and 69 so ACI conversion still works

Related to #620

### DIFF
--- a/contrib/dnsmasq/Dockerfile
+++ b/contrib/dnsmasq/Dockerfile
@@ -1,6 +1,6 @@
-FROM alpine:3.5
+FROM alpine:3.6
 MAINTAINER Dalton Hubble <dalton.hubble@coreos.com>
 RUN apk -U add dnsmasq curl
 COPY tftpboot /var/lib/tftpboot
-EXPOSE 53
+EXPOSE 53 67 69
 ENTRYPOINT ["/usr/sbin/dnsmasq"]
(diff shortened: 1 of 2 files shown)


dgonyeo:
@vip8439 can you confirm that this is the entire set of the ports that would be useful to be exported here?

dghubble:
We should probably note that ACI isn't a focus for us anymore. Maybe there are some migration docs we can link to? @philips

vip8439:
@dgonyeo If I use --net=host instead, those are the complete list of ports dnsmasq is bound to. @dghubble I do believe some migration docs would be helpful if any are available. Should I be running dnsmasq via docker instead? If so, I can look that up.

dghubble:
There are rkt and docker docs specifically for dnsmasq [here](https://github.com/coreos/matchbox/tree/master/contrib/dnsmasq) and I'll poke some people to see if there  are docs on the broader recommendations in the general case.

vip8439:
@dghubble Thanks. Honestly, I am new to Container Linux and made an incorrect assumption that rkt had replaced Docker completely on Container Linux. I saw docker instructions before, but assumed they were for those who wanted to run the image on other distributions and had docker installed instead of rkt. Now I see that Container Linux has both docker and rkt installed.

PR #628, opened 2017-08-04
Fix swallowed errors in server package tests

alrs:
Two error variables were being swallowed unhandled in the tests for the server package.


dghubble:
Thanks

PR #629, opened 2017-08-05
Documentation: remove ETCDCTL_API=3 settings

heyitsanthony:
etcd examples set ETCDCTL_API=3 but are using v2 etcdctl commands. This
works on CL by accident because it ships with 2.3 so etcdctl doesn't
recognize the API env var.


dghubble:
Thanks

PR #635, opened 2017-08-09
matchbox/client: Validate client endpoint is a host:port

### LINKED ISSUE #633: client package should validate gRPC endpoint is a host:port
`grpc.Dial(target string, opts...)` expects a `host:port` target per their [example client](https://github.com/grpc/grpc-go/blob/master/examples/route_guide/client/client.go#L42).

Related #630

dghubble:
* Provide better errors to clients that forget to specify the port or include a protocol scheme by mistake
* grpc-go uses net.SplitHostPort to validate server listener addresses are 'host:port' but doesn't validate Dial targets. https://github.com/grpc/grpc-go/blob/e81b5698fd6a1fdbae0c62f52e9aabdeee231b99/clientconn.go#L304
* Validating our clients provide a host:port is a simple sanity check / pass filter, it is by no means complete. Properly validating a client target depends on the resolver, see https://github.com/grpc/grpc-go/issues/1346.

Before, in a client like `terraform-provider-matchbox`, invalid endpoints produced:

```
failed to create Matchbox client or connect to https://matchbox.example.com:8081: context deadline exceeded
```

Its not very helpful if you don't know the format is host:port. Once this change is pulled in, the messages will be more like:

```
failed to create Matchbox client or connect to matchbox.example.com: client: invalid host:port endpoint: missing port in address matchbox.example.com
failed to create Matchbox client or connect to https://matchbox.example.com8081: client: invalid host:port endpoint: too many colons in address https://matchbox.example.com:8081
```

cc @cehoffman
Closes #633

### DIFF
--- a/matchbox/client/client.go
+++ b/matchbox/client/client.go
@@ -3,6 +3,8 @@ package client
 import (
 	"crypto/tls"
 	"errors"
+	"fmt"
+	"net"
 	"time"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


cehoffman:
Much more useful message. I would have known to look at the endpoint more carefully.

PR #636, opened 2017-08-10
Add version.txt download on get-coreos

jcmoraisjr:
The `version.txt` file is used by coreos-install if the version number is `current`.


dghubble:
Sure

Issue #637, opened 2017-08-11
"(23) Failed writing body" on get-coreos


vitkhab:
When I'm running /matchbox/scripts/get-coreos inside Parallels VM (Alpine Linux) I consistently getting error
```
alpine36:/matchbox# ./scripts/get-coreos
(23) Failed writing body
Channel or Version not found
```

It might be reproduced inside alpine container
```
docker run --rm -t -i alpine:latest /bin/sh
apk add --no-cache git bash curl
git clone https://github.com/coreos/matchbox.git
```
And rerunning `get-coreos` script several times (stopped every successful run by ctrl+c)
```
/ # /matchbox/scripts/get-coreos
Downloading CoreOS stable 1409.7.0 images and sigs to //examples/assets/coreos/1409.7.0
CoreOS Image Signing Key
######################################################################## 100.0%
/matchbox/scripts/get-coreos: line 43: /usr/bin/gpg: No such file or directory
coreos_production_pxe.vmlinuz...
#                                                                          1.5%^C
/ # /matchbox/scripts/get-coreos
(23) Failed writing body
Channel or Version not found
```
Seems stressing cpu helps to reproduce the problem. For stressing I used
```
for i in {0..15}; do yes > /dev/null & ; done
# running get-coreos
killall yes
```

dghubble:
Its just a bash script. What about running `curl` on its own?

dghubble:
You can get the images from https://stable.release.core-os.net/amd64-usr/current/

dghubble:
I think this can be closed

PR #638, opened 2017-08-11
Fix for "(23) Failed writing body" on get-coreos

vitkhab:
Fix for #637 

Curl expects output stream to be open, but `grep -q` closes it right after it found a match.
Solution is to read all the curl output before passing it to grep.


dghubble:
Any idea why this is specific to alpine? There was a suggestion that curl has a flag that handles this in a nicer way that the double pipe, might be worth considering.

PR #644, opened 2017-08-22
libvirt: don't pass --pxe

squeed:
In virt-install v1.4.2, the meaning of  `--pxe` changed from "allow pxe boot" to "always pxe boot." This breaks matchbox, since we expect hosts to pxe-boot only with empty hds. On hosts with v1.4.2, the VMs loop, re-installing CL over and over.

The flag isn't necessary anyways, since we pass `--boot=hd,network`, which enables pxe-booting.

[This](https://github.com/virt-manager/virt-manager/commit/ff3b4dc5b0b21393dbccc0f5f691b17bf1a761bd) is the offending virt-manager commit.

### DIFF
--- a/Documentation/grub.md
+++ b/Documentation/grub.md
@@ -62,5 +62,5 @@ $ sudo docker run --rm --cap-add=NET_ADMIN quay.io/coreos/dnsmasq -d -q --dhcp-r
 Create a VM to verify the machine network boots.
 
 ```sh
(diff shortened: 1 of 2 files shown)


lucab:
Double-checked on v1.4.0 (currently shipped by debian/ubuntu) and this seems not to add regressions there.

dghubble:
Good catch, thanks! I've got `virt-install --version 1.4.1` on Fedora 25 and hadn't noticed yet. This change doesn't add any regressions there either.

PR #645, opened 2017-09-01
Add support for Calico networking

dghubble:


dghubble:
We should probably just use the module in https://typhoon.psdn.io/bare-metal/, its going to be better maintained over time and adds features faster. Its not so practical for me to keep updating both.

dghubble:
Support for flannel or calico networking is included in #646

PR #649, opened 2017-09-17
matchbox/client: Expose Select endpoint

radhus:
Exposes the Select endpoint in matchbox/client just as the other
endpoints like Profiles, Ignition and Generic.

I might have missed if there is a reason for this not to be exposed, but I wanted the ability to lookup a profile based on e.g. a MAC-address (like the /grub HTTP API does).


dghubble:
Yep, its exposed by the gRPC. There's no reason the client can't make it available to callers, there just wasn't a use case before and it was overlooked.

dghubble:
Thanks

radhus:
Thanks for merging, this is where I will be using it FYI https://github.com/radhus/xenmatchbox :)

PR #646, opened 2017-09-12
examples: Update Kubernetes from v1.7.3 to v1.7.5

dghubble:
* Update bootkube version from v0.6.1 to v0.6.2
* Update manual Kubernetes cluster example from v1.7.3 to v1.7.5
* Update Terraform Kubernetes cluster example from v1.7.3 to v1.7.5
   * Switch Terraform example to use Typhoon project's module instead of a copy in this repo
   * Includes support for Calico and Flannel

PR #650, opened 2017-09-18
bootkube calico support move assets_dir/manifests-networking to assets_dir/manifests

bcg62:
As noted in bootkube-terraform, this step must be done for new networking support

https://github.com/poseidon/bootkube-terraform/commit/a52f99e8cc8b395cf2b28f74a9f79c01b63e99ae


dghubble:
In #646, I actually updated the example usage to pull directly from the poseidon/typhoon module, which provides the same options as the examples/terraform/modules/bootkube module. The advantage being I don't have to maintain modules in two places and I think they're a good example use case.

The examples/terraform/modules/bootkube module and the examples/terraform/profiles (that use bootkube) aren't called by the documented examples anymore. A few lines up from your change, notice these files still set Kubernetes v1.7.3 while the documented examples have moved on to v1.7.5. I was planning to remove them and assist with any issues. Can you describe your setup?

bcg62:
Ah, great. I didn't know they were in the other location. No problem for me, removing them would help eliminate confusion. Thanks.

dghubble:
Yes, I'll remove the unused modules in this repo so the example is more clear.

The one change I'll call out is that the module in the poseidon project uses calico as the default. In the example here, just to match how the old example works, I've set the default to flannel. https://github.com/coreos/matchbox/blob/master/examples/terraform/bootkube-install/variables.tf#L68. Feel free to choose either.

PR #652, opened 2017-09-20
Add gpxe boot support.

fudanchii:
Hi, wondering if anyone interested in gPXE boot support for matchbox.
We are using Libvirt KVM in our environment and it has gpxe by default.

Actually we tried with gpxe -> ipxe chainloading but apparently http download performance from this ipxe is totally subpar compared to using gpxe directly.

Of course we'd like to hear if anyone has experience with this kind of setup.

And this patch is basically a copy-paste of the ipxe endpoint code, would love some input.

Thank you!


dghubble:
gPXE is no longer developed, you should try to switch to iPXE. At a glance, this is identical to the iPXE endpoint, except its at `/gpxe`. The endpoint name shouldn't really matter, just use the iPXE endpoint.

fudanchii:
Hi @dghubble 

Yeah we tried to use gpxe boot with matchbox ipxe endpoint at first, but apparently gpxe wont parse pxe script with `#!ipxe` header and will just get hung. I also changed the template so it returns script with `#!gpxe` header, hence the new endpoint.

But yeah, we still finding out why http download from chainloaded ipxe is really slow, and will looking to change the default qemu pxe to use ipxe natively in oppose to chainloading gpxe to ipxe.

dghubble:
QEMU/KVM uses iPXE by default and has for as long as I can remember on my distro (currently Fedora 26, v2.9.0). What is your host system like?

Concerning the slower downloads, what iPXE image are you chainloading? Does the VM have multiple NICs and perhaps they don't both select the same NIC? What's the device model of the interface(s).

I'm not keen to add a gPXE endpoint because we're not in a good position to maintain it. We'd need to go find an older distro and environment that still supports it. Our efforts may be better spent reporting any issues to upstream iPXE.

fudanchii:
Apparently we're unfortunate enough that some of our system still stuck with libvirt-0.10.2 in CentOS 6.7. And I can confirm iPXE is the default for libvirt/Qemu in CentOS 7.3.

We use this [undionly.kpxe](http://boot.ipxe.org/undionly.kpxe) from ipxe.org, and the VM only have 1 NIC attached to it.

I think, for the time being, we will stay with this patch for the existing old systems. The newer systems works OK with iPXE. so we will looking forward to retire these old systems gradually.

I'm going to close this, and thank you for the concern.

PR #659, opened 2017-10-05
Documentation: minor documentation changes

rlenferink:
- Updated coreos versions
- Added code for copying certificates for later use to ~/.matchbox/
- Updated matchbox example host within docker run command


dghubble:
Can you drop the Container Linux version change part? The rest is good, thanks!

We have a (oh so beautiful) sed process that bumps the Container Linux version throughout the project that we run every so often. Its important that the examples and docs all match up so no newcomer caches the wrong version and is confused when the tutorial doesn't work.

rlenferink:
@dghubble i just deleted the version changes from my commit. The oh so beautiful sed can do it's work right now ;)

PR #662, opened 2017-10-21
Add hosts-entry=host to kubelet-wrapper rkt ags

dghubble:
* Kubernetes v1.8.x hyperkube no longer has its own /etc/hosts
* Fixes potential port-forward issues on networks which do not resolve localhost for nodes
* See https://github.com/kubernetes-incubator/bootkube/issues/740

### DIFF
--- a/examples/ignition/bootkube-worker.yaml
+++ b/examples/ignition/bootkube-worker.yaml
@@ -41,7 +41,8 @@ systemd:
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
           --mount volume=opt-cni-bin,target=/opt/cni/bin \
           --volume var-log,kind=host,source=/var/log \
-          --mount volume=var-log,target=/var/log"
+          --mount volume=var-log,target=/var/log \
+          --hosts-entry=host"
         ExecStartPre=/bin/mkdir -p /opt/cni/bin
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/manifests
         ExecStartPre=/bin/mkdir -p /etc/kubernetes/cni/net.d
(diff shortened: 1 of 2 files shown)


examples/ignition/bootkube-worker.yaml
```diff
@@ -41,7 +41,8 @@ systemd:
           --volume opt-cni-bin,kind=host,source=/opt/cni/bin \
           --mount volume=opt-cni-bin,target=/opt/cni/bin \
           --volume var-log,kind=host,source=/var/log \
-          --mount volume=var-log,target=/var/log"
+          --mount volume=var-log,target=/var/log \
+          --hosts-entry=host"
```
lucab: This will start failing as soon as the wrapping script introduces the same flag: https://github.com/rkt/rkt/blob/053f014c38bf0408f68533837bd457034a907afb/rkt/run.go#L650

/cc @squeed

dghubble: Hm, but until that lands some clusters need this to have functional `port-forward` :(

dghubble:
On bare-metal, this flag is needed to fix `port-forward` in 1.8.x hyperkube's only if the upstream DNS isn't   resolving localhost for clients. In most bare-metal environments, this probably isn't an issue. GCE is the only one I know of that doesn't.

Given the choice between "fixing" now but failing those clusters when a new Container Linux release rolls out vs. just pointing those affected to this issue, I agree and favor the later.

dghubble:
Maybe re-open if https://github.com/rkt/rkt/pull/3833 merges.

dghubble:
I'm going to close this for now. The only platform I've found where this change is really really needed is Google Cloud. On bare-metal, it will depend on the network, but generally I think DNS servers handle localhost.

Issue #248, opened 2016-06-23
Respect Ignition Headers


dghubble:
When multiple different Ignition 2.0.0+ versions are available, respect the Ignition headers in order to determine the highest version which can be used when requesting Fuze YAML conversion.

For example, a CoreOS instance at Ignition format 2.1.0 can accept Ignition 2.0.0, but it is preferable to convert a Fuze config to 2.1.0 for that machine by calling fuze `config.ParseAs2_1_0`.

dghubble:
After Ignition v2.1.0 is released, we'd like to start work on this. Assigning code work to @dgonyeo and we can discuss the design/plan together. cc @crawford

dghubble:
`ct` doesn't have configurable versions so this is moot. Instead, we're just dropping support for Container Linux versions below 1465.0.0 in Matchbox v0.7.0.

PR #670, opened 2017-11-06
Update Ignition config version to v2.1.0

### LINKED ISSUE #668: systemd.units.`enabled` is not seen in /ignition JSON
Using matchbox 0.6.1, it seems that `enabled` is not working when fetching JSON from /ignition. Changing it back to 'enable' works fine.

If I create following in ignition profile yaml:

```
systemd:
  units:
    - name: test.service
      enabled: true
      contents: |
        [Unit]
        Description=true
        [Service]
        Type=oneshot
        RemainAfterExit=true
        ExecStart=/bin/true
        [Install]
        WantedBy=multi-user.target
```

then there is no field `enabled` when fetching JSON `/ignition` endpoint:

```
curl -s 'http://192.168.121.156:8080/ignition?uuid=9d087fdf-f8a5-44f6-8ca7-a6ad9b44bbf8&mac=52-54-00-f7-78-09' | jq '.systemd.units|map(select(.name=="test.service"))[0]'
{
  "name": "test.service",
  "contents": "[Unit]\nDescription=true\n[Service]\nType=oneshot\nRemainAfterExit=true\nExecStart=/bin/true\n[Install]\nWantedBy=multi-user.target\n"
}
```

should I change it to `enable`, it then appears:
```
{
  "name": "test.service",
  "enable": true,
  "contents": "[Unit]\nDescription=true\n[Service]\nType=oneshot\nRemainAfterExit=true\nExecStart=/bin/true\n[Install]\nWantedBy=multi-user.target\n"
}
```

Reason I tried to change it  from `enable` to `enabled`, is following warning in ignition output:
```
Nov 03 12:03:37 localhost ignition[369]: files: the enable field has been deprecated in favor of enabled
```
,but it seems that it can't be done with current version of matchbox

dghubble:
* Allows users to write Container Linux Configs with (Ignition) v2.1.0 features.
  * Container Linux Config format https://github.com/coreos/container-linux-config-transpiler/blob/master/doc/configuration.md (v2.1.0 features not explicitly made clear)
* Drop support for Container Linux versions below 1465.0.0 (breaking)
* Update Container Linux config transpiler to v0.5.0
* Update Ignition to v0.19.0, render v2.1.0 Ignition configs

Initially there was a plan that Ignition would request a config and pass the highest version it could accept via headers. It turns out, `ct` only supports conversion to a specific format (which we're now bumping to v2.1.0 so people can use those features) and I'm unwilling to do multiple vendoring hackery.

We will cut a Matchbox v0.7.0 release soon.

Abandons: #248
Closes #668

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,9 @@ Notable changes between releases.
 ## Latest
 
 * Add gRPC API endpoints for managing generic (experimental) templates
+* Update Container Linux config transpiler to v0.5.0
+* Update Ignition to v0.19.0, render v2.1.0 Ignition configs
+* Drop support for Container Linux versions below 1465.0.0 (breaking)
 * Remove Profile `Cmdline` map (deprecated in v0.5.0), use `Args` slice instead
 * Remove pixiecore support (deprecated in v0.5.0)
 * Remove `ContextHandler`, `ContextHandlerFunc`, and `NewHandler` from the `matchbox/http` package.
--- a/glide.lock
+++ b/glide.lock
@@ -1,16 +1,16 @@
-hash: 9c2d24d482e4b591552c20d251b4620bdc2a12256564f8f18d7cb30b797bfd33
-updated: 2017-08-08T13:27:35.465951914-07:00
+hash: b404b094b7ff5d83fac658393148a51f2b3f74ce1026502524be71772c30e9b2
+updated: 2017-11-06T13:24:02.819805752-08:00
 imports:
 - name: github.com/ajeddeloh/go-json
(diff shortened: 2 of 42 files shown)


dghubble:
That Jenkins job is defunct, but continues posting build status. Fixing.

ahjohannessen:
Please push a docker image too :)

dghubble:
Master images are supposed to be auto-published to quay with the appropriate SHA, to compliment official releases. Looks like this broke recently #671 

Thanks for pointing it out.

dghubble:
https://quay.io/repository/coreos/matchbox?tab=tags

PR #671, opened 2017-11-08
travis.yml: Fix travis to publish master images

dghubble:
Travis CI is supposed to auto-publish images from master with the SHA tag to compliment official releases. In #647, the build and deploy matrix stopped matching so publishing stopped. This should fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,6 @@ deploy:
   skip_cleanup: true
   on:
     branch: master
-    go: '1.8.3'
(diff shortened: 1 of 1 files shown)


dghubble:
This job seems to indicate it would start publishing again once merged to master. https://travis-ci.org/coreos/matchbox/jobs/299357700

dghubble:
`Skipping a deployment with the script provider because this is not on the required runtime` - nope

dghubble:
See #672

PR #674, opened 2017-11-16
contrib/dnsmasq: Add ipxe.efi for dnsmasq:v0.5.0

dghubble:
* Add ipxe.efi to dnsmasq image's`/var/lib/tftpboot` directory
* Build `quay.io/coreos/dnsmasq:v0.5.0` Docker image
* Add initrd kernel argument respected only by UEFI https://github.com/coreos/bugs/issues/1239
* Improve network-setup docs and scripts to cover UEFI clients and to support launching UEFI QEMU/KVM clusters locally
* Ensure `./libvirt create-uefi` is usable with cluster examples
* Reduce references to grub.efi flow, its not a happy path

rel: https://github.com/coreos/bugs/issues/1239

### DIFF
--- a/contrib/dnsmasq/docker0.conf
+++ b/contrib/dnsmasq/docker0.conf
@@ -11,8 +11,22 @@ dhcp-host=52:54:00:d7:99:c7,172.17.0.24,1h
 enable-tftp
 tftp-root=/var/lib/tftpboot
 
+# Legacy PXE
+dhcp-match=set:bios,option:client-arch,0
+dhcp-boot=tag:bios,undionly.kpxe
+
+# UEFI
+dhcp-match=set:efi32,option:client-arch,6
+dhcp-boot=tag:efi32,ipxe.efi
+
+dhcp-match=set:efibc,option:client-arch,7
+dhcp-boot=tag:efibc,ipxe.efi
+
+dhcp-match=set:efi64,option:client-arch,9
+dhcp-boot=tag:efi64,ipxe.efi
+
+# iPXE
 dhcp-userclass=set:ipxe,iPXE
-dhcp-boot=tag:#ipxe,undionly.kpxe
 dhcp-boot=tag:ipxe,http://matchbox.foo:8080/boot.ipxe
 
 log-queries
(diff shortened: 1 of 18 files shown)


dghubble:
All the Terraform-based example need to be updated as well to work with UEFI clients.

dghubble:
PXE-enabled DHCP and docker/rkt quay.io/coreos/dnsmasq images show how to run a network boot environment which supports both BIOS and various UEFI clients at the same time.

I don't currently have a dnsmasq proxy-DHCP network environment available anymore, so I've left those examples as-is, they only show the flags for BIOS support.

rockwyc992:
LGTM, but have this be tested this on bare metal machine?

dghubble:
@rockwyc992 I appreciate that, but no, this is still a WIP and won't yet work on Terraform-based examples or simulated disk installs, let alone in the lab.

contrib/dnsmasq/docker0.conf
```diff
@@ -11,8 +11,22 @@ dhcp-host=52:54:00:d7:99:c7,172.17.0.24,1h
...
+dhcp-match=set:efibc,option:client-arch,7
+dhcp-boot=tag:efibc,ipxe.efi
+
+dhcp-match=set:efi64,option:client-arch,9
+dhcp-boot=tag:efi64,ipxe.efi
+
+# iPXE
 dhcp-userclass=set:ipxe,iPXE
-dhcp-boot=tag:#ipxe,undionly.kpxe
 dhcp-boot=tag:ipxe,http://matchbox.foo:8080/boot.ipxe
```
adarshaj: s/matchbox.foo/matchbox.example.com/

dghubble: I believe I originally generated the fake certs with `matchbox.foo` as the name, and these confs are actually used for the locally runnable examples.

dghubble: Actually nope. I believe we can safely change this now.

```
    X509v3 Subject Alternative Name: 
                DNS:matchbox.example.com, DNS:matchbox.foo
```

dghubble:
Proxy-DHCP still needs to be addressed and real-word testing is needed in followups that I'll link here. However, existing functionality is not impacted.

contrib/dnsmasq/docker0.conf
dghubble: Oops. This rename broke all the cluster examples. Fix in #676

PR #675, opened 2017-11-20
fix loading multiple initrds

redbaron:
iPXE loads multiple initrds when each is given to it's own initrd command


dghubble:
Was this tested with two images?

redbaron:
Yes, see https://github.com/coreos/bugs/issues/1862#issuecomment-345791216

dghubble:
Thanks!

PR #681, opened 2017-12-12
examples: Fix examples to work with Terraform v0.11.x

dghubble:
* Explicitly pass provider modules to satisfy constraints
* https://github.com/hashicorp/terraform/issues/16824


dghubble:
I've verified this locally. Without this change, you'd have issues deleting resources created by Terraform after upgrading to v0.11.x, its a rather nasty change in the provider/module inheritance structure inside Terraform.

PR #693, opened 2018-01-25
Update bootkube render; apiserver is now port 6443

colemickens:


dghubble:
These docs instruct users to install bootkube v0.9.1, before bootkube changed to 6443. I would say we just fix this as part of updating to bootkube v0.10.0, but v0.10.0 includes other breaking changes I don't wish to adopt.

colemickens:
Ah, sorry about that. Thanks for the detail.

dghubble:
Its unfortunate kubelet TLS bootstrap wasn't an opt-in. This will be an issue going forward with bootkube. cc @ericchiang

dghubble:
Opened #694 to track the broader issue.

PR #696, opened 2018-01-25
docs: change links to work with sync

zbwright:


dghubble:
Neither of these seem to work:

https://coreos.com/matchbox/docs/latest/terraform/simple-install/
https://coreos.com/matchbox/docs/latest/terraform/simple-install

zbwright:
sync process should translate them to https://github.com/coreos/matchbox/tree/master/examples/terraform/simple-install/ - it'll point down into the repo. these pages aren't linked on the site. 

let me find an example ...

Issue #694, opened 2018-01-25
Update to bootkube v0.10 and beyond


dghubble:
bootkube v0.10.0 changed a few aspects of the Kubernetes setup.

* Changes apiserver port from 443 to 6443 (do-able here)
* Switched kubelet TLS to use TLS bootstrapping

Its unfortunate the later change wasn't an opt-in. As is, kubelet TLS bootstrap changes the secret distribution steps and the admin kubeconfig, but does not improve security. I've posted a longer comment [here](https://github.com/poseidon/terraform-render-bootkube/issues/41).

This needs to be resolved to continue maintaining bare-metal examples.

rel: https://github.com/coreos/matchbox/pull/693
cc: @colemickens @ericchiang

dghubble:
Closing due to #759

PR #703, opened 2018-04-04
Update terraform Kubernetes examples to v1.10.0

dghubble:
cc @sdemos

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ Notable changes between releases.
 
 ## Latest
 
+* Upgrade Kubernetes example clusters to v1.10.0 (Terraform-based)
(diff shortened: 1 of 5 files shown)


sdemos:
Changes look good, I'm testing them out now. 

We probably don't want to let the bootkube-sans-terraform instructions be on a different version than the bootkube-terraform instructions. Is it possible to get those instructions updated too? I can just do it in a separate pr after this if you don't want to.

dghubble:
Non-terraform setup is stalled on https://github.com/coreos/matchbox/issues/694

sdemos:
ah, right. that's unfortunate. 

I successfully tested the changes locally and everything works as expected with terraform.

PR #720, opened 2018-09-24
Add kernel args variable

salarmgh:
Add the `kernel_args` variable in order to pass additional arguments to the kernel. For example, one can pass `coreos.autologin` to the kernel for debugging.


dghubble:
LGTM, although longer term, it may be better for the examples not to bother with variables and just show an example module inputs directly.

PR #726, opened 2018-10-29
scripts: add helper script `get-flatcar`

schu:
Similar to `get-coreos`, add a helper script `get-flatcar` to download
Flatcar assets.

Follow up for https://github.com/poseidon/typhoon/pull/315


dghubble:
LGTM

Issue #729, opened 2019-01-14
Document for getting started (Error: invalid or unknown key: raw_ignition)


rockwyc992:
I used the example in [matchbox/examples/terraform/bootkube-install](https://github.com/coreos/matchbox/tree/master/examples/terraform/bootkube-install), and I found that `terraform-provider-matchbox` have added a new feature for `raw_ignition ` ([reference here](https://github.com/coreos/terraform-provider-matchbox/releases/tag/v0.2.0))

Please update the [document](https://github.com/coreos/matchbox/blob/master/Documentation/getting-started.md#terraform) of getting-started at least v0.2.0 for `terraform-provider-matchbox`, or you will get an error like this
`Error: module.cluster.matchbox_profile.workers[1]: : invalid or unknown key: raw_ignition`

dghubble:
You're right, it requires v0.2.0. Terraform v0.10 had a way to enforce this, but v0.11 lost that ability so you see the error. I'll copy over some newer docs. Although that particular example is basically just https://typhoon.psdn.io/cl/bare-metal/#terraform-setup

dghubble:
https://github.com/coreos/matchbox/pull/731 should clear this up

PR #731, opened 2019-01-24
Improve Terraform tutorial and update example to Kubernetes v1.13.2

dghubble:
* Define the required versions of terraform-provider-matchbox and terraform-provider-ct, with install instructions
* Refresh the tutorial for using Matchbox to PXE boot local QEMU/KVM machines (using Terraform as a client)
* Update Container Linux from 1911.4.0 to 1967.3.0


dghubble:
Should resolve https://github.com/coreos/matchbox/issues/729

PR #740, opened 2019-03-11
Render Container Linux Configs as Ignition v2.2.0, update ignition and config-transpiler

### LINKED ISSUE #715: render v2.2.0 Ignition configs
We need to have access to:
ignition : security : tls : certificateAuthorities

Which was added to the v2.2.0 ignition config spec that is now marked stable.

This is so that we are able to use self-signed certificates as we can't find a way for ignition to allow for them.
e.g. _curl --insecure_ or _wget --no-check-certificate_

I see matchbox-0.7.0 just implemented v2.1.0.  Can you provide how far out on the roadmap when v2.2.0 will be supported?

Thanks!

dghubble:
* Update container-linux-config-transpiler module from v0.5.0 to v0.9.0
  * Render Container Linux Configs referenced in Profiles as Ignition v2.2.0 documents.
* Update ignition module from v0.22.0 to v0.31.0
  * Serve raw Ignition configs (ending in .ign/.ignition) with v2.1 or v2.2 formats (previously, configs above v2.1 produced warnings that the config was too new)

Closes #715


dghubble:
Testing with quay.io/dghubble/matchbox:v0.7.0-62-g5bfcb24-dirty

Issue #712, opened 2018-08-25
dnsmasq image out of date, containing security issues - most likely many others as well


shoeper:
The dnsmasq image has been last build in novemver 2017, older tags even earlier. The dependency alpine:3.6 has been updated several times in the meantime, thus all tags need a rebuild to fix the security vulnerabilities.

Most likely this also effects many other images provided in here.

robinsondan87:
It would also be beneficial to us to upgrade dnsmasq image to something that is newer, we see random ipxe hanging issues which apparently are resolved in later versions of dnsmasq

dghubble:
As per the coreos-user's repo move announcement, the dnsmasq image will now be published under `quay.io/poseidon/dnsmasq` along with Matchbox.

There is a new mechanism to build it via Travis from pushes to a `dnsmasq` branch, but tag images by the repo SHA from which they were produced (no overriding). 

* https://quay.io/repository/poseidon/dnsmasq?tab=tags

dghubble:
Related: 

* https://github.com/poseidon/matchbox/pull/742
* https://github.com/poseidon/matchbox/commit/f4623c508ff3fbc467285de1ede61126624b91ac

shoeper:
The image needs to be rebuild when the base image has been updated as well. Otherwise security issues are going to accumulate.

dghubble:
The based image was already bumped, it's in the links above.

shoeper:
That's not enough. The base image can still change and in that case this image needs a rebuild as well.

dghubble:
I know what your point is, but I'm not planning to take on automated builds.

Prior coreos/dnsmasq images were built by me manually. Now there's a nicer way to bump them when there are significant bumps to alpine. Its sufficient to power the Matchbox examples which use the dnsmasq image in local QEMU/KVM.

PR #742, opened 2019-03-13
Update coreos/dnsmasq image and automate deploys

dghubble:
* Automatically publish dnsmasq image on merge to a branch named "dnsmasq". Similar to how matchbox is automatically published from master. Set the version based on git SHA
* Update base image from alpine:3.6 to alpine:3.9 to address #712
* https://docs.travis-ci.com/user/deployment/script/#passing-arguments-to-the-script (most similar)

Followups:
Need push access and if everything is working, `dnsmasq` branch should probably be protected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,18 @@ install:
 script:
   - make
 deploy:
-  provider: script
-  script: scripts/dev/travis-docker-push
-  skip_cleanup: true
-  on:
-    branch: master
-    go: '1.11.7'
+  - provider: script
+    script: scripts/dev/travis-docker-push
+    skip_cleanup: true
+    on:
+      branch: master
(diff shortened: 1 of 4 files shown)


dghubble:
Hm, Travis says it won't deploy (since branch name is not `dnsmasq`), but later tries anyway :thinking:

dghubble:
Now it seems ok. I think what happened is that I did also have a PR proposing to merge into `dnsmasq` [here](https://github.com/coreos/matchbox/pull/741) (which shows it does build the dnsmasq image and try to push it, failure on creds is expected at this time). These PRs had the same commit hash for a time and I've seen this confuse CI systems. Pushed a timestamp refreshed change and it re-runs afresh and correctly decides that neither deploy step should be run yet.

PR #759, opened 2019-07-08
Remove Kubernetes cluster provisioning examples

dghubble:
* Matchbox examples should be simple and educational to show how to PXE provision machines into clusters. Today, these goals are achieved well enough by the 3-node etcd
cluster example
* Several years ago, I put together the examples PXE booting Kubernetes clusters with Matchbox. That was before we wrote Tectonic or Kubernetes was as popular as it is. Today, a Kubernetes distro is a project in its own right. It no longer makes sense to maintain (duplicate) a Kubernetes distro as "an example" inside Matchbox.
* Matchbox is now used for Kubernetes cluster provisioning in more organizations than ever. It backs the poseidon/Typhoon and kinvolk/Locomotive distros. These both serve as great external examples of using Matchbox to provision Kubernetes clusters

Attention: If you relied on Matchbox Kubernetes docs, you can find a similar guide at https://typhoon.psdn.io/cl/bare-metal/ (same author). https://github.com/poseidon/typhoon/


dghubble:
Closes https://github.com/poseidon/matchbox/issues/719 https://github.com/poseidon/matchbox/issues/694

PR #762, opened 2019-09-05
WIP: latest ignition 3 support

spaced:
this is only a WIP, but like to discuss about support of different ignition variants.


dghubble:
I'm not sure if you're aware, but the Matchbox API accepts prepared Ignition documents directly or CLC configs templatized at render/serve time (historical). I recommend the former mechanism, it is advantageous to use template and/or transpiler mechanisms _during_ the preparation of the Ignition document (i.e. as easrly as possible). For example, to show template errors in their original context, to show error lines in the original source files before merge into a single document, to configure strictness at edit time rather than at serve time, etc.). As opposed to my Matchbox's own templating mechanism that I wrote before such nice external tools were in place.

Matchbox can already be used to serve Ignition v2.2 or Ignition v3.0 today and supports Container Linux and Fedora CoreOS - see [terraform-provider-ct](https://github.com/poseidon/terraform-provider-ct) and more broadly, Typhoon, for Terraform examples. For other configuration systems, it would be best to adopt a similar strategy.

I'm not very keen to extend Matchbox's own templating feature further. There are real advantages to using the native features/mechanisms/approaches of configuration systems (at least in Terraform-land) to prepare and validate the Ignition document before sending it to Matchbox.

PR #768, opened 2019-10-30
Document uefi chainloading in proxy mode

rkoster:
As discussed [here](https://forum.ipxe.org/showthread.php?tid=10746)


dghubble:
Ah nice, good to see use of the embedded [ipxe.efi](https://github.com/poseidon/matchbox/blob/master/contrib/dnsmasq/get-tftp-files#L13). One note about those you probably noticed, those "official" blobs from ipxe folks aren't compiled with TLS support.

dghubble:
Thanks :clinking_glasses:

PR #782, opened 2020-01-16
Reload gRPC TLS cert and key on every connection

sedan07:
This enables automatic rotation of the cert/key without requiring restarting of the daemon. Such as when running in a container and using short life certs from automated PKI service.

It does however reload the cert and key from disk on every new connection. On RPC requests this should make very little difference as this is as an admin endpoint and generally would not be under constant load.


dghubble:
I'd want this to be done via the usual sighup signal approach.

PR #783, opened 2020-02-08
Update get-fedora-coreos convenience script

dghubble:
* Refer to Fedora CoreOS 31 download paths
* https://github.com/poseidon/typhoon/pull/624

### DIFF
--- a/scripts/get-fedora-coreos
+++ b/scripts/get-fedora-coreos
@@ -4,8 +4,8 @@
 #
 set -eou pipefail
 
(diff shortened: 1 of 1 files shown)


lucab:
Just FYI:
 * there is a containerized installer/downloader for FCOS at https://github.com/coreos/coreos-installer
 * there is JSON metadata for all channels/releases, see https://github.com/coreos/fedora-coreos-tracker/tree/master/metadata

PR #785, opened 2020-02-11
Added powerPC support at .travis.yml file

marcelo-martins:
Adding support to Power architecture by including "ppc64le" in the arch section of travis.yml.
Currently, we have the binary for Matchbox-v0.8.3 stored at our FTP server.
https://oplab9.parqtec.unicamp.br/pub/ppc64el/matchbox/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: go
 sudo: required
+arch:
+  - amd64
+  - ppc64le
 services:
   - docker
 go:


dghubble:
What's your intention exactly? If you're wanting releases to include a pre-compiled binary for a new architecture, you probably want to propose adding it at https://github.com/poseidon/matchbox/blob/master/Makefile#L102

The binaries for alternative architectures are mostly for conveneince. I'm doubtful that running the tests on additional architectures is really worth it time-wise or would be acted upon if it failed (if a less common architecture couldn't easily be made to align with amd64 it probably gets dropped).

marcelo-martins:
You are right. I am going to follow your suggestion.

PR #787, opened 2020-02-19
Added ppc64le support

marcelo-martins:
This adds ppc64le(power architecture) support to the Makefile, as suggested by @dghubble


dghubble:
Thanks

PR #788, opened 2020-03-22
`contrib/k8s` not in the release images

barakmich:
The docs suggest it is (https://matchbox.psdn.io/deployment/#kubernetes) but they aren't there. They're only in the git checkout. Which is fine, but for consistency.


dghubble:
Thanks

PR #791, opened 2020-04-25
Fix for Kubernetes

ssmiller25:
Two minor fixes to allow Kubernetes deployments to work:

* Move the deploy to the app/v1 api version ([removed from Kubernetes 1.16 onward](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/) - the app/v1 has been supported since Kubernetes 1.9)
* Modify the RPC ingress name so that is does not conflict with the primary matchbox ingress.

### DIFF
--- a/contrib/k8s/matchbox-deployment.yaml
+++ b/contrib/k8s/matchbox-deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: matchbox
@@ -7,11 +7,15 @@ spec:
   strategy:
     rollingUpdate:
       maxUnavailable: 1
+  selector:
+      matchLabels:
+        app: matchbox
   template:
     metadata:
       labels:
         name: matchbox
         phase: prod
+        app: matchbox
     spec:
       containers:
         - name: matchbox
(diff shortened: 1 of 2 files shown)


contrib/k8s/matchbox-ingress.yaml
```diff
@@ -15,7 +15,7 @@ spec:
 apiVersion: extensions/v1beta1
```
dghubble, line 15: While you're updating, mind changing these both to networking.k8s.io/v1beta1

contrib/k8s/matchbox-deployment.yaml
```diff
@@ -7,11 +7,15 @@ spec:
   strategy:
     rollingUpdate:
       maxUnavailable: 1
+  selector:
+      matchLabels:
+        app: matchbox
```
dghubble, line 12: match `name: matchbox` please

contrib/k8s/matchbox-deployment.yaml
```diff
@@ -7,11 +7,15 @@ spec:
...
   template:
     metadata:
       labels:
         name: matchbox
         phase: prod
+        app: matchbox
```
dghubble, line 18: let's stick with just name please

contrib/k8s/matchbox-ingress.yaml
ssmiller25: Not a problem - and done

contrib/k8s/matchbox-deployment.yaml
ssmiller25: Sounds good.  Does make the selector/labels a bit cleaner.

dghubble:
thanks

PR #804, opened 2020-06-16
Fix simple-install to work with terraform 12

spuder:
With terraform v0.12.13, this gives the following errors
```
Error: Unsupported block type

  on groups.tf line 18, in resource "matchbox_group" "node1":
  18:   selector {

Blocks of type "selector" are not expected here. Did you mean to define
argument "selector"? If so, use the equals sign to assign it a value.
```


dghubble:
Examples still target Terraform v0.11 and this is v0.12 specific. However, some time before next release, examples need to be either updated or removed so this is fine.

dghubble:
Thanks

PR #806, opened 2020-06-19
Add Ubuntu example

spuder:
Adds 2 examples installing ubuntu 18.04. One using 'assets' cacheing and one using archive.ubuntu.com.

## Limitations
Due to the way that preseed works, it's not possible to put the 'hostname' inside a preseed config. It _must_ be passed in as a kernel option, or be derived from dhcp. https://askubuntu.com/a/675620/172065

```
  args = [
   ....
    "netcfg/get_domain=unassigned-domain",
    "netcfg/get_hostname=unassigned-hostname",
]
```

I've tested many options, and can't find a way to use a variable inside `args=[]`

e.g

```
args = [
"netcfg/get_domain=$${hostname}",
]
```
If it _is_ possible to do so, please let me know.


dghubble:
Hey Spencer, I appreciate the spirit here.

Matchbox focuses on modern container-optimized, declarative, Ignition-based OSes (Flatcar Linux, Fedora CoreOS). As you've noticed, its possible to use Matchbox with traditional OS distros (e.g. cloud-init and kickstart). But I don't want to encourage these usages or give false hope that these OSes are a target use case or that I'd test and maintain Ubuntu examples.

I appreciate you sharing your work though. It may be a helpful pointer to someone. I think this might best live in your own examples or elsewhere you mention.

PR #812, opened 2020-10-09
Improvement - Adding Sprig to Matchbox Templating

DandyDeveloper:
I don't know if you're still taking contributions or if this is considered interesting, however, in the templating we've (My org) needed to include Sprig to make some of the templating better defined and simpler. 

This PR is just an option if you were interested in including this with the official Matchbox binary. 

If not, feel free to close!


DandyDeveloper:
@dghubble Just pinging you for this because I'm not sure who is taking charge here right now.

dghubble:
No thanks. Matchbox templating features aren't being expanded upon, with the recommendation of using external components to prepare/template/preprocess configs for addition via the gRPC API.

dghubble:
Also, I see across all projects I run, no need to ping.

n0rad:
I have the same need as @DandyDeveloper . While preparing the config out of matchbox is useful (and I do). There is some use case where you need minimum templating functions to prepare for example the ignition config based on the group using it.

I my case I need to substring a group metadata keys list and loop over it to setup an ignition part.

I don't understand why you did not merged it. It's bringing value while changing only 3 lines and not touching api or storage.

Can you please reconsider merging it ?

PR #825, opened 2020-11-08
Specify arch in get-flatcar script

anthr76:
I'm currently using this on my own as I have a multi arch environment. I have added the `arch` flag at the end making it minimum viable and not a breaking change. amd64 is the default.

Unsure if it's desired to have separate directories dependent on architectures. It's a small change and more work is necessary to make matchbox/typhoon a little more multi-arch friendly. Feel free to toss out if this isn't desired.

### DIFF
--- a/scripts/get-flatcar
+++ b/scripts/get-flatcar
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # USAGE: ./scripts/get-flatcar
-# USAGE: ./scripts/get-flatcar channel version dest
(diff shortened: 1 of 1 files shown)


dghubble:
I've tested Flatcar Linux arm64. But as of this writing, its not ready for real usage and has issues that aren't being resourced upstream. Since the examples and scripts imply some degree of quality, I'd prefer to wait before including Flatcar Linux arm64.

dghubble:
Can revisit if Flatcar Linux arm64 gets to a useful point.

PR #827, opened 2021-01-04
Is Matchbox Dead?

the-maldridge:
This project seems intentionally hostile to use, with no issues available, and precious little documentation.  The gRPC example no longer appears to work either, and the only way of getting in touch appears to be to open a pull request, which I'm doing here.

So far I've tried creating the certs and starting up everything pointing at the right files, and bootcmd just hangs forever.  When using the terraform provider it does eventually time out due to the context expiring, but it also doesn't do anything useful either.

If the Poseidon Labs/matchbox stack is not actually intended to be used that's fine, I can go find some other software, but this does feel like a bug somewhere in the matchbox code.

### DIFF
--- a/foo
+++ b/foo
@@ -1,0 +1,1 @@
+file intentionally empty


dghubble:
No, the project is in active use.

Github issues are not offered, because folks conflate open-source with support, help, advice, and operational assistance. Be sure to read the getting started tutorials and if you have a contribution you can make a PR.

the-maldridge:
Thanks for the info.  I'll move along to another solution given that the provided TLS script fails to even complete successfully in the examples, much less getting as far as booting a machine or anything that approaches a production quality deployment.

Were there issues I'd be happy to share my findings, but that would constitute me supporting your project, offering you advice on improvement, or otherwise providing operational assistance to people continuing to use matchbox, which you have made clear the project is not interested in.

Best of luck in your endeavors!

PR #875, opened 2021-08-13
Improvement : Matchbox learns --files-dir

5bentz:
Hello,

I have noticed container-linux-config-transpiler (ct) [supports including local files](https://github.com/coreos/container-linux-config-transpiler/blob/e4d5be564a0bda8197080222ebcf0a184143061c/doc/configuration.md) in the configuration, for example :
```
storage:
  files:
    - filesystem: "root"
      path: "/etc/test"
      contents:
        local: test
```

However, with matchbox, it would yield this error : the `ct` library/exec complains about a missing opt-arg.

```
time="2021-08-13T13:44:43Z" level=error msg="error converting Container Linux config: error at line 185, column 16\nlocal files require setting the --files-dir flag to the directory that contains the file"
```

Therefore my PR. I hardly have any knowledge in Go and I have been inspiring myself from the [clc code base](https://github.com/coreos/container-linux-config-transpiler/blob/e4d5be564a0bda8197080222ebcf0a184143061c/internal/main.go#L35)

Please, let me know what you think about it. Is this flag support desirable?

Regards,

### DIFF
--- a/cmd/matchbox/main.go
+++ b/cmd/matchbox/main.go
@@ -29,6 +29,7 @@ func main() {
 		rpcAddress   string
 		dataPath     string
 		assetsPath   string
+        filesDir     string
 		logLevel     string
 		grpcCAFile   string
 		grpcCertFile string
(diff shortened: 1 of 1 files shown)


dghubble:
Butane Config `local` fields are specific to the butane command line tool. I don't wish to see this in Matchbox.

Matchbox's recommended style of usage is to render/merge Ignition contents externally and submit them to the Matchbox API (e.g. https://github.com/poseidon/terraform-provider-ct). Matchbox doing its own templating of CLC configs is an older feature I'd like to remove before v1, and don't wish to expand upon. Having Matchbox interpreset `local` to mean some random location on the Matchbox server seems further astray. Plus, if this were added, the next thing folks would want is an API to manage those files remotely. Its too complicated. Better tools exist to prepare/render Ignition,  Matchbox's role is just to serve it (along with iPXE, etc).

https://github.com/coreos/fedora-coreos-tracker/issues/375
https://github.com/poseidon/terraform-provider-ct/pull/68

5bentz:
It makes total sense having separate tools for rendering and serving.
Thanks a lot for your explanation.

PR #890, opened 2021-10-11
Implement Container Linux config spec 3.x support via butane transpiler

icedream:
Flatcar and Fedora CoreOS use different major versions of the Container Linux configuration specifications, with each of them developing their own versions of the respective parser.

I ran into trouble using a YAML file to provide Ignition configuration for Fedora CoreOS. It turns out things such as `storage.directores` and `storage.files` no longer needing `filesystem: root` are incompatibilities with the CoreOS Container Linux configuration parser that is still used in Matchbox.

I could not simply replace the transpiler with the one used for Fedora CoreOS (butane) since [it does not support Ignition specs older than 3.0.0](https://github.com/coreos/fedora-coreos-tracker/blob/main/Design.md#supported-ignition-versions), also explicitly laid down by [this comment](https://github.com/coreos/fedora-coreos-tracker/issues/89#issuecomment-485966453). Flatcar on the other hand continues to use the 2.x specs.

As such I have thought about how to handle the situation and decided to split how the configuration is parsed based on whether `version`/`variant` fields have been specified.

- If no version is defined, 2.x or older spec for flatcar is assumed, so the flatcar transpiler is used.
- If a version is defined but `variant` is `flatcar`, also use the flatcar transpiler. This is for future-proofing this solution and/or provide an escape hatch if they ever decide to use a variation of the 3.x spec.
- In all other cases Fedora CoreOS transpiler will be used.

Marking this as draft until I have fully tested this out with currently existing versions of Flatcar and Fedora CoreOS. Not testing against old legacy CoreOS, this is most likely going to be breaking change for that case (Flatcar defaults to 2.3.x spec, old CoreOS transpiler outputs 2.2.x Ignition configs).


dghubble:
Using Fedora CoreOS Butane (formerly fcct) or CoreOS Config is already supported through the Ignition API, with tools like https://github.com/poseidon/terraform-provider-ct that handle transpiling. We're not extending Matchbox's early templating system.

Duplicate of https://github.com/poseidon/matchbox/pull/762

PR #923, opened 2022-02-18
Allow passing alternative architecture for assets download

ohmer:
The scripts can download assets for ARM too.
Tested with:

- `./scripts/get-fedora-coreos stable 35.20220131.3.0 . aarch64`
- `./scripts/get-flatcar "" "" . arm64`


dghubble:
Thanks for this, but since the local QEMU/KVM examples don't show ARM64, this seems somewhat unneccessary. If anything, I'm aiming to make the examples simpler.

PR #977, opened 2022-07-26
Update iPXE example to match the FCOS iPXE documentation

jdoss:
This PR updates the examples to match the FCOS iPXE [docs](https://docs.fedoraproject.org/en-US/fedora-coreos/bare-metal/#_installing_from_ipxe). Without these changes, the server will fail to boot with this error:

```
/dev/root can't open blockdev
```


dghubble:
Thanks. https://github.com/poseidon/matchbox/pull/978 has more context on _why_ this is needed on UEFI

PR #993, opened 2022-09-11
Update ignition version

duhruh:
I should start by saying, this change fixed the issue I was running into so i'm not sure of the full ramifications of this change. I kept hitting [this error](https://github.com/poseidon/matchbox/blob/main/matchbox/http/ignition.go#L67
) when attempting to follow the flatcar route of https://matchbox.psdn.io/getting-started/ 

It would appear the ignition file generated by terraform is a higher format than allowed by the version of ignition parser used. Updating the ignition package to the latest version and using the `v3_3` parser worked for me.


dghubble:
Flatcar Linux only recently added support for Ignition v3.3 in newer images. This will be part of the larger migration from Container Linux Configs (rendered as Ignition 2.2) to Butane/Ignition v3.3 happening across projects gradually (its breaking and needs other supporting work)

Rel:

* https://github.com/poseidon/terraform-provider-ct/pull/132
* https://github.com/poseidon/typhoon/pull/1196

dghubble:
Also, the warning in the logs should be ignorable.

duhruh:
In my case this warning meant that the ignition file was ignored completely, at least in my testing. The OS was installed but none of my ignition file changes were applied.

dghubble:
You may be interested in the examples. Both Fedora CoreOS and Flatcar Linux use Ignition v3.3 in them and they've been validated recently: https://github.com/poseidon/matchbox/pull/980

duhruh:
Hey sorry, i guess there's some misunderstanding. What i'm saying is I did follow those tutorials/examples and ran into these issues. Only after I did a manual build/deploy of matchbox with this change did that flatcar example work.

dghubble:
I was never able to repro the issue you were seeing, but Ignition v2 (spec v3.x) is now the only supported version as part of the planned Container Linux Config deprecation. https://github.com/poseidon/matchbox/pull/997 should have the update you were wanting.

PR #1001, opened 2022-10-01
Configureable architecture for get-flatcar script

toanju:
This introduces a fourth parameter to the get-flatcar script which allows to select the architecture of the downloads. This allows to download files for arm64. The options for this parameter are 'amd64 and arm64. The default is unchanged 'amd64'.

Example:
./get-flatcar beta current ./lib/assets arm64


dghubble:
Duplicate of #923. The project doesn't really need to maintain a generalized downloader script

toanju:
Would have been nice for my Raspberry Pis, but then I'll keep it on my side.

PR #1056, opened 2023-02-24
[storage] add etcd as a backend

mgrzybek:
This PR allows Matchbox to store data (profiles, groups, ignition files) into an ETCD KV store. As a consequence, the service can be run in a stateless way.

* The storage type (file or etcd) can be specified using `--storage-type`.
* The endpoints can be specified using `--etcd-endpoints`.
* The keys are prefixed using their type (profiles, groups, ignitions) in the store.

### DIFF
--- a/cmd/matchbox/main.go
+++ b/cmd/matchbox/main.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


dghubble:
Thanks, but I don't really have a use case where it's worth it to me to run an etcd for this (vs just a persistent volume). So I'd rather not maintain or validate this to provide mainline support.

PR #1069, opened 2023-03-28
CPU architecture detection with iPXE

Sadi-a:
Some commands were added in the ipxeBootstrap to enable architecture detection based on the architecture the iPXE version used was built for. 
This is done through the use of an additional flag in the ipxeBootstrap string using the buildarch parameter.


dghubble:
Seems non-harmful, sure

PR #1077, opened 2023-04-05
examples: repair variable expansion for mac and uuid in examples

jklare:
* currently there is a duplicated $ in front of some of the variables used to render the ignition_url mac and uuid matches, leading to broken variable expansion and impossible matches
* remove duplicated $ to for all ignition_urls and update docs accordingly

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -107,7 +107,7 @@ resource "matchbox_profile" "fedora-coreos-install" {
     "initrd=main",
     "coreos.live.rootfs_url=https://builds.coreos.fedoraproject.org/prod/streams/${var.os_stream}/builds/${var.os_version}/x86_64/fedora-coreos-${var.os_version}-live-rootfs.x86_64.img",
     "coreos.inst.install_dev=/dev/sda",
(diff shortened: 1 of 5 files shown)


dghubble:
These examples define Matchbox resources in Terraform, which uses `$`  to interpolate Terraform variables such as `${var.matchbox_http_endpoint}` at apply time. Matchbox profile kernel arguments are used to populate iPXE configs, use parameters like [`${mac:hexhyp}`](https://ipxe.org/cfg/mac) which is evaluated at boot time on each matched machine and requires double escaping.

The examples are correct already. You can find more examples in https://github.com/poseidon/typhoon/blob/main/bare-metal/fedora-coreos/kubernetes/profiles.tf#L11

jklare:
I probably was a bit too quick on patching too many files. I think the double $$ are needed for the .tf files, but is this syntax also correct for the .json ones ? I used them as given and ran into issues while booting since the variables where not correctly expanded. I can post the issues tomorrow morning once I am back in the office. Thanks for the quick review :)

dghubble:
There is one json example with a double escape that shouldn't be there. Fixed in https://github.com/poseidon/matchbox/pull/1080

PR #1264, opened 2024-05-27
Fix asset fetching scripts and complain earlier in case of failure

eldondev:
I was using the get-flatcar script and I noticed a failure as a result of some changes in the layout of the assets. I updated the scripts to fail if they encounter a server error.


.github/workflows/test-script.yaml
```diff
@@ -0,0 +1,12 @@
+name: test-example-scripts
```
dghubble, line 1: Can you drop this? I don't really care to have tests for the helper script since it's just provided as a convenience.

eldondev: Yep. Done.

PR #1457, opened 2025-08-02
scripts/get-fedora-coreos: Fix kernel URL

samcday:


samcday:
I wonder if this script even needs to exist at all nowadays? Perhaps it would make more sense to drop it and update docs/references to use `coreos-installer` instead? If you agree LMK and I'll open a follow-up PR for that.

PR #1551, opened 2026-03-03
Add -L flag to curl in scripts/get-flatcar to follow redirects.

bugre:
Flatcar downloads from '[channel].flatcar-linux.net' now redirect (302) to 'flatcar.cdn.cncf.io/[channel]', so to still point the base url to flatcar's own domain, adding the '-L' ( '--location' ) will allow the script correctly get the needed files.

---

> _$ curl -sI https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt
HTTP/2 302
alt-svc: h3=":443"; ma=2592000
cache-control: max-age=300
location: https://flatcar.cdn.cncf.io/stable/amd64-usr/4459.2.3/version.txt
server: Caddy
date: Tue, 03 Mar 2026 20:40:04 GMT_

### DIFF
--- a/scripts/get-flatcar
+++ b/scripts/get-flatcar
@@ -39,37 +39,37 @@ fi
 echo "Downloading Flatcar Linux $CHANNEL $VERSION images and sigs to $DEST"
 
 echo "Flatcar Linux Image Signing Key"
-curl -f# https://www.flatcar.org/security/image-signing-key/Flatcar_Image_Signing_Key.asc -o "${DEST}/Flatcar_Image_Signing_Key.asc"
+curl -L -f# https://www.flatcar.org/security/image-signing-key/Flatcar_Image_Signing_Key.asc -o "${DEST}/Flatcar_Image_Signing_Key.asc"
 $GPG --import <"$DEST/Flatcar_Image_Signing_Key.asc" || true
 
 # Version
 echo "version.txt"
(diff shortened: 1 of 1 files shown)


bjwschaap:
Hi @bugre I actually was about to create an issue + PR for the exact same thing. Also, it would be nice to add an $ARCH variable/parameter to be able to select aarch64/arm64 next to the x64/amd64. If you like I can put my version in a Gist as a reference, instead of us both creating a PR for the same thing...

bjwschaap:
@bugre 
- https://gist.github.com/bjwschaap/934dde419c4bde54a3d64191e8542589
- https://gist.github.com/bjwschaap/0ac4cc5de8a18bc43387c2349fe4aa21

bugre:
Hi @bjwschaap,
Adding the support for a optional "ARCH" flag to the two 'get' scripts isn't much related to the fix of curl following redirects i would suggest a separated PR, as usually maintainers prefer not mixing different issues in the same PR ( @dghubble ) ?.

I'll be away for a few days, when i'm back i could pull the upstream changes into my fork and try adding your diff to it. 
I did a quick check and it seems that the second gist:
> named [get-fedora-coreos](https://gist.github.com/bjwschaap/0ac4cc5de8a18bc43387c2349fe4aa21#file-get-fedora-coreos),  actually also has the content of 'get-flatcar'.

